PR #71, opened 2017-03-20
Feature/model fixes

luckytommy:
Remove valueType from ProgramTrackedEntityAttribute
TrackedEntityAttribute.optionSet field typo


luckytommy:
Added sortOrder to ProgramTrackedEntityAttribute and went through the schema. The only thing that might be of interest to us is the field programTrackedEntityAttributeGroups which we might need in the future (for searching).

I also set DataElement as a mandatory foreign key to TrackedEntityDataValue.

PR #72, opened 2017-04-03
Changed UserCall to download paths for OrganisationUnits + version bump

VladislavAlfredov:
This fixes an issue that occurs after, when feeding the retrieved User with assigned OrganisationUnits (with null paths) to OrganisationUnitCall, causing an exception.
Bumped the sdk version.


VladislavAlfredov:
Changes are there

PR #74, opened 2017-04-04
Hotfix/metadata sync transactions

VladislavAlfredov:


core/src/test/java/org/hisp/dhis/android/core/common/MetadataCallTests.java
```diff
@@ -302,20 +317,103 @@ public void setUp() throws IOException {
...
+        verify(databaseAdapter, times(6)).beginNewTransaction();
+        verify(transaction, times(6)).end();
+        verify(transaction, atMost(4)).setSuccessful();
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void call_shouldNotMarkTransactionSuccessful_ifOptionSet_Fails() throws Exception {
+        when(optionSetCall.execute()).thenReturn(errorResponse);
+
```
erlingfjelstad: Create a constant (int transactionsSuccess = 6) so it is not a magic number

PR #79, opened 2017-04-15
Feature/variables

ArazAbishov:
This is the first chunk of implementation of RuleEngine, which primarily includes:
 - APIs of RuleEngine (see readme file as well)
 - Mapping events / enrollments to variables (see RuleVariableMapBuilder). 

Any feedback is welcome. 

P.S. My apologies for huge PR. Up to this point there was no anything worth sending for review.


core-rules/src/main/java/org/hisp/dhis/android/rules/RuleEngine.java
```diff
@@ -1,112 +1,125 @@
...
 
     @Nonnull
-    public List<Event> events() {
-        return events;
+    public Callable<List<RuleEffect>> evaluate(@Nonnull RuleEvent ruleEvent) {
+        if (ruleEvent == null) {
+            throw new IllegalArgumentException("ruleEvent == null");
+        }
+
+        for (RuleEvent contextualEvent : ruleEvents) {
```
ArazAbishov: There is a possibility that client will send ruleEvent with different fields to evaluate(), so we can't rely on .equals() or .hashcode() in this case.

core-rules/src/main/java/org/hisp/dhis/android/rules/RuleVariableValue.java
```diff
@@ -0,0 +1,45 @@
...
+abstract class RuleVariableValue {
+
+    @Nullable
+    abstract String value();
+
+    @Nonnull
+    abstract RuleValueType type();
+
+    @Nonnull
+    abstract List<String> candidates();
```
ArazAbishov: No, because it is just another state of the same model by design. RuleVariableValue might not contain actual value, but can contain expected type (which comes from Variable). This way we can substitute default or "null" value for primitives which can be false, 0 or '' depending on source type.

core-rules/src/main/java/org/hisp/dhis/android/rules/RuleVariableValueMapBuilder.java
```diff
@@ -0,0 +1,397 @@
...
+        buildEnvironmentVariables(valueMap);
+
+        // set metadata variables
+        buildRuleVariableValues(valueMap);
+
+        // do not let outer world to alter variable value map
+        return Collections.unmodifiableMap(valueMap);
+    }
+
+    private static boolean isEventInList(@Nonnull List<RuleEvent> ruleEvents,
```
ArazAbishov: Same reasoning as for .evaluate().

PR #119, opened 2018-01-05
aggregate-generics: include all generic classes built for aggregates

luisyng:
All generic classes implemented for the aggregated data module in a separate Pull Request including: 
- Generic handlers
- Generic stores
- Generic calls
- Generic tests
- Test utilities

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/common/BaseIdentifiableObject.java
+++ b/core/src/main/java/org/hisp/dhis/android/core/common/BaseIdentifiableObject.java
@@ -32,6 +32,7 @@ import android.support.annotation.Nullable;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.text.ParseException;
 import java.util.Date;
 
 public abstract class BaseIdentifiableObject implements IdentifiableObject {
@@ -39,7 +40,7 @@ public abstract class BaseIdentifiableObject implements IdentifiableObject {
     within models which extend BaseIdentifiableObject */
     public static final SafeDateFormat DATE_FORMAT = new SafeDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
 
-    protected static final String UID = "id";
+    public static final String UID = "id";
     protected static final String CODE = "code";
     protected static final String NAME = "name";
     protected static final String DISPLAY_NAME = "displayName";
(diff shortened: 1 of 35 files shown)


core/src/main/java/org/hisp/dhis/android/core/option/OptionSetModel.java
```diff
@@ -38,23 +39,44 @@
...
+import org.hisp.dhis.android.core.common.StatementBinder;
 import org.hisp.dhis.android.core.common.ValueType;
 import org.hisp.dhis.android.core.data.database.DbValueTypeColumnAdapter;
+import org.hisp.dhis.android.core.utils.Utils;
 
+import static org.hisp.dhis.android.core.utils.StoreUtils.sqLiteBind;
 
 @AutoValue
-public abstract class OptionSetModel extends BaseIdentifiableObjectModel {
+public abstract class OptionSetModel extends BaseIdentifiableObjectModel implements StatementBinder {
```
dhis2-android: This class is not generic right? Also, shouldn't it belong to the Tracker domain?

luisyng: The class is not generic, but it contains the additional properties that we have added to the model classes in the aggregate domain to be able to use them with our generic classes. To be able to write integration tests for our generic classes, we had to pick one model class and add these additional properties (in the best case a simple one), because the table should exist in the system during the execution of the integration test. 

It doesn't have any implications in performance or whatsoever, but if you don't want us to touch any class from the Tracker domain, we can leave it out of this pull request and include the tests using DataSet in the next pull request.

core/src/main/java/org/hisp/dhis/android/core/common/BaseIdentifiableObject.java
```diff
@@ -32,14 +32,15 @@
...
+import java.text.ParseException;
 import java.util.Date;
 
 public abstract class BaseIdentifiableObject implements IdentifiableObject {
     /* date format which should be used for all Date instances
     within models which extend BaseIdentifiableObject */
     public static final SafeDateFormat DATE_FORMAT = new SafeDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
 
-    protected static final String UID = "id";
+    public static final String UID = "id";
```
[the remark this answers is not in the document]
luisyng: Because we build generic procedures for Identifiable and Nameable objects and we need for those the name of the columns that each one provides.

PR #123, opened 2018-01-13
Aggregate datasets

marcamsn:
DataSet and DataSetDataElement support.


dhis2-android:
We are missing n * n relationships between org. units and datasets

PR #124, opened 2018-01-19
Migration from deprecated android-apt to annotationProcessor

Ezmeyers:
These changes would allow updating the Android Gradle plugin to version 3.0.1 and Gradle to version 4.1

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -30,7 +30,6 @@
 def libs = rootProject.ext.libraries
 
 apply plugin: 'com.android.library'
-apply plugin: 'com.neenbedankt.android-apt'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
@@ -79,7 +78,7 @@ dependencies {
     compile 'com.squareup:otto:1.3.6'
 
     // Raizlabs libraries.
-    apt 'com.raizlabs.android:DBFlow-Compiler:2.2.1'
+    annotationProcessor 'com.raizlabs.android:DBFlow-Compiler:2.2.1'
     compile 'com.raizlabs.android:DBFlow-Core:2.2.1'
     compile 'com.raizlabs.android:DBFlow:2.2.1'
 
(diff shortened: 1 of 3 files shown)


ui/build.gradle
```diff
@@ -42,6 +42,7 @@ android {
         versionCode cfg.versionCode
         versionName cfg.versionName
         buildConfigField "String", "main_app", '"'+cfg.package+'"'
+        vectorDrawables.useSupportLibrary = true
```
ifoche: Just like with dagger comment, I think you're using vectorDrawables in your project and you need this. Please if this is something that only affects your project, can you remove it from here? Thanks!

core/build.gradle
```diff
@@ -115,4 +114,7 @@ dependencies {
 
     testCompile "junit:junit:${libs.jUnitVersion}"
     testCompile 'commons-logging:commons-logging:1.1.1'
+
+    compile 'com.google.dagger:dagger:2.0'
```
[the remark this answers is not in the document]
Ezmeyers: I will remove this, you are correct.

ui/build.gradle
Ezmeyers: When updating, you are asked to "add google maven repository and sync project," after doing so, without the above line, a wall of these errors occurs:

dhis2-android-sdk/ui/src/main/res/drawable/ic_arrow_back.xml : Can't process attribute android:fillColor="@android:color/white": references to other resources are not supported by build-time PNG generation. See http://developer.android.com/tools/help/vector-asset-studio.html for details.

This occurs with the official dhis2-tracker-capture and not our local project

ifoche: @Ezmeyers you're right. This is needed to fully complete the upgrade keeping backward compatibility to some properties that we're using in some xmls. I'm accepting this change. BTW, following your PR we've decided to upgrade versions of most of the libraries we're using. We'll introduce it for the next version. Thanks for the contribution!

PR #125, opened 2018-01-20
Sdk aggregated

dhis2-android:


dhis2-android:
Adding aggregate domain to data model (still no datavalues)

PR #147, opened 2018-02-23
Periods and datavalues

marcamsn:
Periods and DataValues.

Solves [ANDROSDK-106](https://jira.dhis2.org/browse/ANDROSDK-106) and [ANDROSDK-19](https://jira.dhis2.org/browse/ANDROSDK-19)


luisyng:
We have updated the PR with Jose's Feedback. Aggregated Data is now downloaded in a separate AggregatedDataCall which can be directly triggered in D2.

PR #150, opened 2018-02-28
value-type-image: add value type image

luisyng:
MetadataCallRealIntegrationShould was complaining because the value IMAGE was missing in the enum, probably introduced in 2.29.


luisyng:
Hi! This is already in master because other PR were on top of it. Should we decline it?

dhis2-android:
Yes. I think, we don't need this PR. It has been already introduced

PR #183, opened 2018-04-13
Program refactor

luisyng:
- Refactors the program block with the generic architecture. 
- Moves program related endpoint calls to `ProgramParentCall`. 
- Adds `GenericEndpointCallImpl` an exception throwing when response is failed to prevent calls that chain endpoint calls have to check individually if they are successful.


core/src/main/java/org/hisp/dhis/android/core/program/ProgramStage.java
```diff
@@ -76,40 +76,40 @@
...
+    private static final NestedField<ProgramStage, ProgramStageSection> programStageSections
             = NestedField.create(PROGRAM_STAGE_SECTIONS);
-    public static final NestedField<ProgramStage, ProgramStageDataElement> programStageDataElements =
+    private static final NestedField<ProgramStage, ProgramStageDataElement> programStageDataElements =
             NestedField.create(PROGRAM_STAGE_DATA_ELEMENTS);
-    public static final NestedField<ProgramStage, ObjectStyle> style = NestedField.create(STYLE);
-    public static final NestedField<ProgramStage, ObjectWithUid> program = NestedField.create(PROGRAM);
-    public static final NestedField<ProgramStage, Access> access = NestedField.create(ACCESS);
+    private static final NestedField<ProgramStage, ObjectStyle> style = NestedField.create(STYLE);
+    private static final NestedField<ProgramStage, ObjectWithUid> program = NestedField.create(PROGRAM);
```
luisyng: I have used ObjectWithUid just for the NestedField but not for the actual property to prevent static initialization errors, since Program requires PogramStage and ProgramStage requires Program.

PR #201, opened 2018-04-30
System info date

luisyng:
Solves [ANDROSDK-141](https://jira.dhis2.org/browse/ANDROSDK-141).


core/src/main/java/org/hisp/dhis/android/core/trackedentity/TeisEndPointCall.java
```diff
@@ -84,6 +83,9 @@ private TeisEndPointCall(@NonNull TrackedEntityInstanceService trackedEntityInst
                     transaction.end();
                 }
             }
+
+            resourceHandler.handleResource(ResourceModel.Type.TRACKED_ENTITY_INSTANCE,
```
luisyng: Handler wasn't being updated here.

PR #203, opened 2018-04-30
Tei search response pojo

luisyng:
Tei search response pojo (SearchGrid) and tests. Transformation into TrackedEntityInstance will be done on top. @vgarciabnz, @josemp10 please say if terminology has to be changed.


vgarciabnz:
About terminology, dhis2-core uses Grid and GridHeader in [`common` package](https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common). And ListGrid as the implementation in [`support`](https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid).

Anyway we can refactor them in another PR

PR #206, opened 2018-05-03
android-sdk-28

luisyng:
Upgrades Android Target SDK to API 27 (Android 8.1). It doesn't affect to the minimum SDK version supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ android:
   components:
   - tools
   - platform-tools
-  - build-tools-25.0.2
-  - android-25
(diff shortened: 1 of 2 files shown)


luisyng:
Okay, Travis is failing with that. I thought that would be automatic. We can just decline it if that's going to make us lose any time.

```
 You have not accepted the license agreements of the following SDK components:
     [Android SDK Platform 27].
     Before building your project, you need to accept the license agreements and complete the installation of the missing components using the Android Studio SDK Manager.
```

PR #265, opened 2018-06-13
Uid splitter refactor

vgarciabnz:
Refactors [ANDROSDK-177](https://jira.dhis2.org/browse/ANDROSDK-177)

Solves:
- [ANDROSDK-236](https://jira.dhis2.org/browse/ANDROSDK-236)
- [ANDROSDK-237](https://jira.dhis2.org/browse/ANDROSDK-237)
- [ANDROSDK-239](https://jira.dhis2.org/browse/ANDROSDK-239)


core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityTypeCall.java
```diff
@@ -128,6 +128,8 @@
 
     public static final UidsCallFactory<TrackedEntityType> FACTORY = new UidsCallFactory<TrackedEntityType>() {
 
+        private static final int MAX_UID_LIST_SIZE = 140;
```
luisyng: This is not used

PR #322, opened 2018-07-23
Program stage description

marcamsn:
Solves [ANDROSDK-293](https://jira.dhis2.org/browse/ANDROSDK-293).


core/src/androidTest/java/org/hisp/dhis/android/core/program/ProgramStageModelShould.java
```diff
@@ -74,6 +74,7 @@
     private static final PeriodType PERIOD_TYPE = PeriodType.Weekly;
     private static final Integer ACCESS_DATA_WRITE = 1;
     private static final Integer REMIND_COMPLETED = 1;
+    private static final String DESCRIPTION = "description";
```
vgarciabnz: @marcamsn Include displayDescription field, both in model and api object

marcamsn: @vgarciabnz, the displayDescription field has been included.

PR #325, opened 2018-07-25
Aggregated - Add support for Data input periods

AbelCS:
[ANDROSDK-78](https://jira.dhis2.org/browse/ANDROSDK-78)


core/src/test/resources/dataset/data_set.json
```diff
@@ -41,6 +41,15 @@
       }
     }
   ],
+  "dataInputPeriods": [
```
vgarciabnz: Backend model allows to create dataInputPeriods with null openingDate and closingDate. Currently it fails when parsing the object. It should be easily replicated with a test.
Make these two properties nullable.

AbelCS: Ok, I tought they was mandatory. I'll fix it.

AbelCS: Fixed. 

Thanks for your feedback!

PR #408, opened 2018-10-04
category-category-combos-order: add sort order

luisyng:
Solves [ANDROSDK-355](https://jira.dhis2.org/browse/ANDROSDK-355).


luisyng:
The name of the branch is wrong, it should be category-category-combos-order according to the functionality implemented.

PR #418, opened 2018-10-08
Androsdk 365

luisyng:
- Solves [ANDROSDK-365](https://jira.dhis2.org/browse/ANDROSDK-365).
- Reduces constraint in IdentifiableSyncHandlerImpl: only ObjectWithUidInterface is required instead of IdentifiableObject.
- Ensures dataElement stores wiping is done as part of its module and not in the global list in WipeModule


core/src/main/assets/migrations/26.yaml
```diff
@@ -0,0 +1,8 @@
+up:
+- PRAGMA foreign_keys=off;
+- BEGIN TRANSACTION;
+- ALTER TABLE DataElementOperand RENAME TO DataElementOperand_Old;
+- CREATE TABLE DataElementOperand (_id INTEGER PRIMARY KEY AUTOINCREMENT, uid TEXT NOT NULL UNIQUE, dataElement TEXT, categoryOptionCombo TEXT, FOREIGN KEY (dataElement) REFERENCES DataElement (uid) DEFERRABLE INITIALLY DEFERRED, FOREIGN KEY (categoryOptionCombo) REFERENCES CategoryOptionCombo (uid) DEFERRABLE INITIALLY DEFERRED);
+- INSERT INTO DataElementOperand (_id, uid, dataElement, categoryOptionCombo) SELECT _id, uid, dataElement, categoryOptionCombo FROM DataElementOperand_Old;
+- COMMIT;
```
vgarciabnz: @luisyng missing DROP TABLE DataElementOperand_Old?

PR #431, opened 2018-10-15
[ANDROSDK-393] DataValue: Move exists checking from Handler to Store

AbelCS:
[ANDROSDK-393](https://jira.dhis2.org/browse/ANDROSDK-393)


core/src/main/java/org/hisp/dhis/android/core/datavalue/DataValueCollectionRepository.java
```diff
@@ -38,25 +38,27 @@
 final class DataValueCollectionRepository extends ReadOnlyCollectionRepositoryImpl<DataValue>
         implements ReadWriteCollectionRepository<DataValue> {
 
+    private final DataValueStore dataValueStore;
```
vgarciabnz: @AbelCS @luisyng store is private in ReadOnlyCollectionRepositoryImpl. What about making it protected in ReadOnlyCollectionRepositoryImpl so we don't duplicate variables?

luisyng: Yes, that makes a lot of sense.

vgarciabnz: Well, in this case it does not work because it uses the method "exists" in DataValueStore, which is not generic

PR #432, opened 2018-10-15
Androsdk 394

marcamsn:
Solves [ANDROSDK-394](https://jira.dhis2.org/browse/ANDROSDK-394).


core/src/main/java/org/hisp/dhis/android/core/indicator/IndicatorTypeFields.java
```diff
@@ -28,21 +28,28 @@
...
+    final static String FACTOR = "factor";
+
+    private static final FieldsHelper<IndicatorType> fh = new FieldsHelper<>();
+
+    public static final Field<IndicatorType, String> uid = fh.uid();
+
+    static final Field<IndicatorType, String> lastUpdated = fh.lastUpdated();
+
+    public static final Fields<IndicatorType> allFields = Fields.<IndicatorType>builder()
+            .fields(fh.getIdentifiableFields())
```
vgarciabnz: @marcamsn IndicatorType extends BaseNameable

marcamsn: @vgarciabnz The IndicatorType has been left extending IdentifiableObjects and the migration has been done to eliminate the unnecessary columns in the database.

PR #480, opened 2018-11-15
 [ANDROSDK-437] Unified class TrackedEntityDataValue

marcamsn:
Solves [ANDROSDK-437](https://jira.dhis2.org/browse/ANDROSDK-437).


core/src/main/java/org/hisp/dhis/android/core/common/ObjectStoreImpl.java
```diff
@@ -160,7 +160,7 @@ protected int countWhere(@NonNull String whereClause) {
         }
     }
 
-    private void addObjectsToCollection(Cursor cursor, Collection<M> collection) {
+    public void addObjectsToCollection(Cursor cursor, Collection<M> collection) {
```
vgarciabnz: It could be protected better

core/src/main/java/org/hisp/dhis/android/core/common/SQLStatementBuilder.java
```diff
@@ -73,7 +73,7 @@ public SQLStatementBuilder(TableInfo tableInfo) {
                 tableInfo.hasSortOrder());
     }
 
-    private String commaSeparatedColumns() {
+    public String commaSeparatedColumns() {
```
vgarciabnz: It could remain private

PR #699, opened 2019-02-28
[ANDROSDK-690] Unify ProgramSectionAttributeLink

marcamsn:
Solves [ANDROSDK-690](https://jira.dhis2.org/browse/ANDROSDK-690).


core/src/main/java/org/hisp/dhis/android/core/program/ProgramSectionAttributeLink.java
```diff
@@ -25,50 +25,44 @@
...
+import org.hisp.dhis.android.core.common.Model;
 
-@RunWith(JUnit4.class)
-public class ProgramSectionAttributeLinkModelBuilderShould {
+@AutoValue
+public abstract class ProgramSectionAttributeLink implements Model {
 
-    @Mock
-    private ProgramSection programSection;
+    @Nullable
```
vgarciabnz: Could they be NonNull? Just to tell users that a value always exists

PR #803, opened 2019-05-02
SMS branch update

stasheq:
Contains updates done, when really tried to use this functionality in the app.


core/src/main/java/org/hisp/dhis/android/core/sms/data/LocalDbRepositoryImpl.java
```diff
@@ -32,10 +41,16 @@
 
     public LocalDbRepositoryImpl(Context ctx,
```
luisyng: I would recommend you here to add the @Inject annotation and in SmsDIModule: 

```
@Provides
    LocalDbRepository localDbRepository(LocalDbRepositoryImpl) {
        return impl;
    }
```

So you don't need to add the dependencies in both files.

core/src/main/java/org/hisp/dhis/android/core/sms/data/LocalDbRepositoryImpl.java
```diff
@@ -130,4 +133,41 @@ public Completable setMetadataIds(final Metadata metadata) {
                         context.openFileOutput(METADATA_FILE, Context.MODE_PRIVATE), metadata
                 ));
     }
+
+    @Override
+    public Single<Event> getEventToSubmit(String eventUid, String teiUid) {
+        return Single.fromCallable(() ->
```
luisyng: I think this can all be replaced by:

```
Single.fromCallable(() -> eventModule.events.uid(eventUid).withTrackedEntityDataValues())
```

The teiUid should also be set.

core/src/main/java/org/hisp/dhis/android/core/sms/data/LocalDbRepositoryImpl.java
```diff
@@ -130,4 +133,41 @@ public Completable setMetadataIds(final Metadata metadata) {
...
+                eventModule.events.byUid().eq(eventUid).one().get().toBuilder()
+                        .trackedEntityInstance(teiUid)
+                        .trackedEntityDataValues(values)
+                        .build()
+        );
+    }
+
+    @Override
+    public Single<TrackedEntityInstance> getTeiEnrollmentToSubmit(String enrollmentUid, String teiUid) {
+        return Single.fromCallable(() ->
```
luisyng: Can be replaced by: 

```
Single.fromCallable(() -> trackedEntityModule.trackedEntityInstances.uid(teiUid).withEnrollments().withTrackedEntityAttributeValues().get())
```

stasheq:
Requested parts fixed.

PR #1172, opened 2020-03-02
feat: Mock server app

JaimeToca:
The purpose of this PR is to allow the app to use the mockserver from the SDK.  Please add any suggestion or comment to the PR if you think something could be improve :+1: 
Some considerations : 
- A dhis2Dispatcher was created so that could be shared.
- A response controller was also created in order to associate method (post, put, delete, get...), paths, json files and httpcodes. The dispatcher is the one that works with this response controller.
- Since we needed our own implementations of the iFileReader, the interface was used and the implementation is in the app.

The idea was to configure the behaviour of the mock server for each test like this 

`fun shouldLoginSuccessfullyWhenCredentialsAreRight() {
        mockWebServerRobot.addResponse(GET, API_ME_PATH, API_ME_RESPONSE_OK)
        mockWebServerRobot.addResponse(GET, API_SYSTEM_INFO_PATH, 
       API_SYSTEM_INFO_RESPONSE_OK)
       //Test code goes here
  }`

Note: The path constants are defined in the response controller


core/src/main/java/org/hisp/dhis/android/core/mockwebserver/ResponseController.java
```diff
@@ -0,0 +1,80 @@
...
+        codeResponses = new HashMap<>();
+        methodsMap = new HashMap<>();
+        methodsMap.put(GET, new LinkedHashMap<>());
+        methodsMap.put(POST, new LinkedHashMap<>());
+        methodsMap.put(PUT, new LinkedHashMap<>());
+        methodsMap.put(DELETE, new LinkedHashMap<>());
+    }
+
+    void populateInternalResponses(){
+
```
JaimeToca: The idea is to init here methodsMap with the SDK json/paths. If so, I think we could share the new dhis2Dispatcher

JaimeToca:
I'll fix conflicts

core/src/main/java/org/hisp/dhis/android/core/mockwebserver/Dhis2Dispatcher.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class Dhis2Dispatcher extends Dispatcher {
+
+    private static final String DISPATCHER = "Dispatcher";
+
+    private IFileReader fileReader;
+    private ResponseController responseController;
+
+    Dhis2Dispatcher(IFileReader fileReader, ResponseController responseController){
+        this.fileReader = fileReader;
```
JaimeToca: We needed our own implementations of the file reader:
- our jsons sometimes are different than the ones in the sdk/resources
- the current implementation of the IfileReader returns the resources of the sdk
- we work with assets instead

PR #1227, opened 2020-04-10
feat: [androsdk-1007] Set data values state as uploading when syncing

marcamsn:
Solves [ANDROSDK-1007](https://jira.dhis2.org/browse/ANDROSDK-1007).

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/datavalue/internal/DataValuePostCall.java
+++ b/core/src/main/java/org/hisp/dhis/android/core/datavalue/internal/DataValuePostCall.java
@@ -28,19 +28,22 @@
 
 package org.hisp.dhis.android.core.datavalue.internal;
 
+import androidx.annotation.NonNull;
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/hisp/dhis/android/core/datavalue/internal/DataValuePostCall.java
```diff
@@ -74,6 +80,8 @@
                 return systemInfoDownloader.downloadMetadata().andThen(Observable.create(emitter -> {
                     emitter.onNext(progressManager.increaseProgress(SystemInfo.class, false));
 
+                    markPartitionsAsUploading(dataValues);
+
                     DataValueSet dataValueSet = new DataValueSet(dataValues);
 
                     DataValueImportSummary dataValueImportSummary = apiCallExecutor.executeObjectCall(
```
vgarciabnz, line 87: Wrap this part in a try-catch to mark the partition as error in case the response response is an error. Otherwise the partition will remain in UPLOADING state

PR #1229, opened 2020-04-10
feat: [androsdk-1008] Set data set complete registration state as uploading when syncing

marcamsn:
Solves [ANDROSDK-1008](https://jira.dhis2.org/browse/ANDROSDK-1008).


core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationStore.java
```diff
@@ -43,4 +43,6 @@
     boolean removeNotPresentAndSynced(Collection<String> dataSetUids,
                                       Collection<String> periodIds,
                                       String rootOrgunitUid);
+
+    boolean isDSCRBeingUpload(DataSetCompleteRegistration dscr);
```
vgarciabnz, line 47: Better call it isBeingUpload. It is already in the DSCR store

PR #1266, opened 2020-05-19
fix: Fix some spelling

Hamza-ye:
Fix some spelling I could remember that I found while reading.


vgarciabnz:
Thanks @Hamza-ye, these will be visible in the next night build.
If you are working in a app using the sdk and you have any question about it, please reach us in DHIS2 community. Keep in touch!

PR #1268, opened 2020-05-20
feat: [androsdk-1127] [androsdk-1128] Add validation rule support

marcamsn:
Solves [ANDROSDK-1127](https://jira.dhis2.org/browse/ANDROSDK-1127) and [ANDROSDK-1128](https://jira.dhis2.org/browse/ANDROSDK-1128) .


marcamsn:
It will be updated in another PR

PR #1296, opened 2020-06-16
feat: [ANDROSDK-1170] Catch MAX_TEI_COUNT_REACHED error

vgarciabnz:
Solves [ANDROSDK-1170](https://jira.dhis2.org/browse/ANDROSDK-1170)


vgarciabnz:
Postponed to 1.3.0 because this error can't be easily handled by the app

PR #1468, opened 2021-01-29
Merge/support program indicator avg sum types

ifoche:
This PR implements the following functionality:
- avg & sum aggregation type program indicators


vgarciabnz:
Inline program indicators in web application behave differently

PR #1591, opened 2021-07-13
feat: [ANDROSDK-1387] Analytics service, DataElement evaluator

vgarciabnz:
Implement the structure of the AnalyticsService, including the logic to determine the different dimensions and filters. There are evaluators for the different kind of data. DataElement and DataElementOperand operator is implemented so far.

Solves [ANDROSDK-1402](https://jira.dhis2.org/browse/ANDROSDK-1402) and [ANDROSDK-1387](https://jira.dhis2.org/browse/ANDROSDK-1387).


core/src/main/java/org/hisp/dhis/android/core/analytics/aggregated/service/evaluator/AnalyticsEvaluatorHelper.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.hisp.dhis.android.core.period.Period
+import org.hisp.dhis.android.core.period.PeriodTableInfo
+
+internal object AnalyticsEvaluatorHelper {
+
+    const val Sum = "SUM"
+    const val Avg = "AVG"
+    const val Count = "COUNT"
+    const val Max = "MAX"
+    const val Min = "Min"
```
marcamsn, line 46: My intuition tells me that perhaps the Min property should be capitalized.

PR #1596, opened 2021-07-20
[Androapp-4099] Check if enrollment allows event creation

andresmr:


core/src/main/java/org/hisp/dhis/android/core/enrollment/EnrollmentService.kt
```diff
@@ -104,4 +110,30 @@ class EnrollmentService @Inject constructor(
...
+    fun allowEventCreation(enrollmentUid: String, stagesToHide: List<String>): Boolean {
+        val programStages = eventCollectionRepository.byEnrollmentUid().eq(enrollmentUid ?: "")
+            .byDeleted().isFalse.get()
+            .toFlowable().flatMapIterable { events: List<Event>? -> events }
+            .map { event: Event -> event.programStage() }
+            .toList()
+            .flatMap { currentProgramStagesUids: List<String?> ->
+                val repository = programStagesCollectionRepository.byProgramUid().eq(
+                    enrollmentRepository.uid(enrollmentUid).blockingGet().program()
+                )
```
vgarciabnz, line 123: The filter `.byAccessDataWrite().isTrue()` could be use here instead of the validation in line 128

PR #1706, opened 2022-01-24
feat: [ANDROSDK-1472] Visualizations: add legend properties

eric-ampire:
Visualizations have two properties to define how the legends must be rendered


core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/custom/internal/DbVisualizationLegendColumnAdapter.kt
```diff
@@ -0,0 +1,79 @@
...
+        val legendStrategyColumnIndex = cursor.getColumnIndex(VisualizationTableInfo.Columns.LEGEND_STRATEGY)
+        val legendStyleColumnIndex = cursor.getColumnIndex(VisualizationTableInfo.Columns.LEGEND_STYLE)
+        val legendSetIdColumnIndex = cursor.getColumnIndex(VisualizationTableInfo.Columns.LEGEND_SET_ID)
+
+        val legendShowKey = cursor.getString(legendShowKeyColumnIndex)
+        val legendStrategy = cursor.getString(legendStrategyColumnIndex)
+        val legendStyle = cursor.getString(legendStyleColumnIndex)
+        val legendSetId = cursor.getString(legendSetIdColumnIndex)
+
+        val isLegendInfoPresent = listOf(legendShowKey, legendStrategy, legendStyle, legendSetId).all {
```
vgarciabnz, line 51: LegendSetId might be null and still be a valid VisualizationLegend object. Actually, all the properties in VisualizationLegend are nullable, so we could build the object in any case

core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/custom/internal/DbVisualizationLegendSetColumnAdapter.kt
```diff
@@ -0,0 +1,31 @@
+/*
```
vgarciabnz, line 1: Could this class be removed?

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegend.java
```diff
@@ -0,0 +1,91 @@
...
+import org.hisp.dhis.android.core.arch.db.adapters.enums.internal.VisualizationLegendSetColumnAdapter;
+
+
+@AutoValue
+@JsonDeserialize(builder = AutoValue_VisualizationLegend.Builder.class)
+public abstract class VisualizationLegend {
+
+    @Nullable
+    @JsonProperty()
+    @ColumnAdapter(VisualizationLegendSetColumnAdapter.class)
```
vgarciabnz, line 51: I think we don't this column adapter because all the mapping is done in other column adapter.

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegend.java
```diff
@@ -0,0 +1,91 @@
...
+    public abstract VisualizationLegendSet set();
```
vgarciabnz, line 52: This property could be an ObjectWithUid. We only care about the uid.

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegend.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Nullable
+    @JsonProperty()
+    public abstract String showKey();
+
+    @Nullable
+    @JsonProperty()
+    public abstract String strategy();
```
vgarciabnz, line 60: This property could be an enum with two possible values: "BY_DATA_ITEM" and "FIXED" (default).

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegend.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Nullable
+    @JsonProperty()
+    public abstract String style();
```
vgarciabnz, line 64: This property could be an enum with two possible values: "FILL" (default) and "TEXT".

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegendSet.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.auto.value.AutoValue;
+
+
+@AutoValue
+@JsonDeserialize(builder = AutoValue_VisualizationLegendSet.Builder.class)
+public abstract class VisualizationLegendSet {
```
vgarciabnz, line 42: This class could be removed if the LegendSet is an ObjectWithUid

core/src/sharedTest/resources/visualization/visualization.json
```diff
@@ -314,7 +314,12 @@
   "numberType": "VALUE",
   "aggregationType": "DEFAULT",
   "legend": {
-    "hidden": false
+    "showKey": false,
```
vgarciabnz, line 317: We can use this info to verify deserialization in VisualizationShould unit test.

core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/custom/internal/DbVisualizationLegendColumnAdapter.kt
eric-ampire: It's means that the if check is not necessary ?

vgarciabnz: We could remove the check, or replace the "all" by "any" so that if there is at least one property we build the object

core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/custom/internal/DbVisualizationLegendColumnAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val legendShowKey = cursor.getString(legendShowKeyColumnIndex)
+        val legendStrategy = cursor.getString(legendStrategyColumnIndex)
+        val legendStyle = cursor.getString(legendStyleColumnIndex)
+
+        val isLegendInfoPresent = listOf(legendShowKey, legendStrategy, legendStyle).any {
+            it != null
+        }
+
+        return if (isLegendInfoPresent) {
```
vgarciabnz, line 52: Here you have to build the VisualizationLegend object using all the properties, not just legendShowKey. The condition to create is right (if any of them is not null, the object must be created), but the object must contain all the proporties, even if some of them are null.

core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/custom/internal/DbVisualizationLegendColumnAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+                .showKey(legendShowKey)
+                .build()
+        } else {
+            null
+        }
+    }
+
+    override fun toContentValues(values: ContentValues, columnName: String?, value: VisualizationLegend?) {
+        value?.let {
+            values.put(VisualizationTableInfo.Columns.LEGEND_SHOW_KEY, it.showKey())
```
vgarciabnz, line 63: It is the same here, you must persist all the properties

core/src/main/java/org/hisp/dhis/android/core/arch/db/adapters/enums/internal/LegendStrategyColumnAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.db.adapters.enums.internal
+
+import org.hisp.dhis.android.core.visualization.LegendStrategy
+
+class LegendStrategyColumnAdapter : EnumColumnAdapter<LegendStrategy>() {
```
vgarciabnz, line 33: All the classes (adapters) could be "internal"

core/src/main/java/org/hisp/dhis/android/core/visualization/VisualizationLegend.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hisp.dhis.android.core.common.ObjectWithUid;
+
+
+@AutoValue
+@JsonDeserialize(builder = AutoValue_VisualizationLegend.Builder.class)
+public abstract class VisualizationLegend {
+
+    @Nullable
+    @JsonProperty()
+    @ColumnAdapter(ObjectWithUidColumnAdapter.class)
```
vgarciabnz, line 54: I think that all the column adapters in this class could be removed because the handling of this class is done in the DbVisualizationLegendColumnAdapter.
Just to verify that, add a VisualizationLegend property in the sample object "VisualizationSamples". This object is used by the "VisualizationStoreIntegrationShould" test to verify that an object can be persisted and read from the database.

PR #1733, opened 2022-02-28
SDK support for legends in TEI analytics and visualizations analytics

ifoche:
This PR contains what is necessary on the SDK side to show analytics in the app. The list of things implemented here could be summarized as follows:

*TEI analytics*
- Add legend property to LineListResponse
- Add method to configure the repository by legendStrategy (none, byDataItem, fixed)
- Implement calculate of legend for data elements and program indicators in the EventLineListServiceImp according to legendStrategy

*Visualizations analytics*
- Extract common types and logic to reuse code between TEI analytics and visualization analytics
- Add legend property to analytics responses
- Add method to select legend strategy to AnalyticsRepository
- Implement legend for data elements and program indicators calculation in the EventLineListServiceImp according to legendStrategy
- Implement legend strategy in analytics service
- Select analytics legend strategy by visualization legend strategy
- Add new tests

**Recommendations for testing this PR:**
- If this PR is tested without its equivalent in the official Android app, then please ensure to add traces in the ChartCoordinatesProviderImpl, concretely in the indicatorCoordinates and dataElementCoordinates methods, in order to see the configured legends
- In case this PR is tested with the official Android app, please ensure that you're testing in 2.36+ unless the solution to this Jira issue is already in the code: https://jira.dhis2.org/browse/ANDROSDK-1472
- When testing visualizations, please ensure that visualization has the checkbox "Use legend for chart colors" active, as otherwise the app would crash if the variables that are hidden by this checkbox have null value
- When testing pivot tables, please ensure that the visualization only has a single column in the data dimension. The app does not seem to allow multiple columns configurations
- When testing indicators we have found several issues on the calculations of the values, that are completely unrelated to this code and out of the scope of the legends implementation. They are just bugs. Our recommendation is, when testing indicators, to establish the expression directly to the value you want to test (e.g. 2) and to set the filter expression to "true" (the app does not seem to work properly with program indicators that don't contain filter expressions).


core/src/main/java/org/hisp/dhis/android/core/analytics/LegendEvaluator.kt
```diff
@@ -0,0 +1,100 @@
...
+            return legendRepository
+                .byStartValue().smallerThan(value.toDouble())
+                .byEndValue().biggerThan(value.toDouble())
+                .byLegendSet().eq(legendSetUid)
+                .one()
+                .blockingGet() ?: legendRepository
+                .byEndValue().eq(value.toDouble())
+                .byLegendSet().eq(legendSetUid)
+                .one()
+                .blockingGet()
```
vgarciabnz, line 95: The only reason to have two queries instead of just one query is that the operator `biggerOrEqualTo` does not exist in the repo, isn't it? So if we add it we could have just one query

core/src/main/java/org/hisp/dhis/android/core/analytics/aggregated/DimensionalResponse.kt
```diff
@@ -38,5 +40,6 @@ data class DimensionalResponse(
 
 data class DimensionalValue(
     val dimensions: List<String>,
-    val value: String?
+    val value: String?,
+    val legend: Legend?
```
vgarciabnz, line 44: An alternative option would be to have `val legend: String?`, so the legend is not a whole Legend object but just the uid of the Legend. Then, the whole Legend object will be included in the metadata map. This is the strategy followed for `val dimensions: List<String>`, they only contain a list of uids and the metadata map contains the whole objects (name, description, etc). The purpose of this is to reduce the size of the response and avoid having duplicated objects.
I am not sure that this strategy makes sense for Legends becuase it can be very handy to have whole Legend object (including the color) when printing the value in the table. You have implemented this in the App side, so you know better than me. What do you think about this?

vgarciabnz:
I have run the checks and added two commits in a separate branch:
- Ktlint and checkstlye issues: https://github.com/dhis2/dhis2-android-sdk/commit/bc4be23c623d78d5909ae23f3541e77cbe2ece2d
- Supress some warnings in detekt: https://github.com/dhis2/dhis2-android-sdk/commit/8947a7ce8341dd65b0d4c7704dfae8eced4bfa82

You can cherry-pick them to your branch and update the PR. The second commit might be solved by reducing the complexity of the method, but it is fine to me to just suppress the warnings in this case (up to you).

core/src/main/java/org/hisp/dhis/android/core/analytics/LegendEvaluator.kt
ifoche: Thank you @vgarciabnz . We will implement this shortly. We're assuming that you're giving us the green light to modify your ORM, which we need to implement this method.

core/src/main/java/org/hisp/dhis/android/core/analytics/aggregated/DimensionalResponse.kt
ifoche: @vgarciabnz sorry, we understood this was actually what you wanted us to do (misunderstanding probably from our side). In any case, this is also implemented like this in TEI analytics and its modification would (as you properly stated) also imply modifying the app code. Would you mind if we do this as a refactor after this PR, so this can be validated as-is for this release and next one contains the refactor? That would give us the possibility of checking how's the performance impact before moving the code to a new version without duplicated objects

core/src/main/java/org/hisp/dhis/android/core/analytics/LegendEvaluator.kt
vgarciabnz: Great thanks!

core/src/main/java/org/hisp/dhis/android/core/analytics/aggregated/DimensionalResponse.kt
vgarciabnz: I agree on merging the current PR and look at this later. Actually, I am not completely sure about replacing the Legend by just the uid: the benefit of having the whole Legend object when printing the values could be higher that the difference in performance. Just wanted to know if you had considered that option.

vgarciabnz:
There are some conflicts caused by a previous PR that changed how LegendSets are downloaded. There were some changes in the json files used for testing, and these changes were done just a few days ago, so we can easily solve the conflicts in the PR. If it is ok for you, we might handle the conflicts and merge it

vgarciabnz:
Conflicts solved and PR merged here https://github.com/dhis2/dhis2-android-sdk/pull/1743

PR #1782, opened 2022-05-03
feat: [ANDROSDK-1519] Improve perfomence in synchronization of program indicators

eric-ampire:


core/src/main/java/org/hisp/dhis/android/core/program/ProgramIndicator.java
```diff
@@ -61,7 +61,7 @@ public abstract class ProgramIndicator extends BaseNameableObject implements Cor
 
     @Nullable
     @JsonProperty()
-    public abstract String expression();
+    public abstract String  expression();
```
vgarciabnz, line 64: It seems to be a typo

core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramIndicatorCall.kt
```diff
@@ -0,0 +1,91 @@
...
+                    fields = ProgramIndicatorFields.allFields,
+                    displayInForm = null,
+                    program = null,
+                    uids = ProgramIndicatorFields.uid.`in`(partitionUids),
+                    false
+                )
+            }
+        )
+
+        return Single.merge(firstPayload, secondPayload).reduce { t1, t2 ->
```
vgarciabnz, line 85: The handling of the items is done within the `reduce` body. It will work in this case because there are only two singles, but if we add a third one, like `Single.merge(firstPayload, secondPayload, thirdPayload)`, the code will compile and will look ok but the handling will be done multiple 2 times:
- First time (t1 + t2)
- Second time ((t1 + t2) + t3)

It will be better to use other operator to wait for all the calls to complete. Or reduce the items and do the handling outside the reduce function.

core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramIndicatorHandler.kt
```diff
@@ -29,23 +29,42 @@ package org.hisp.dhis.android.core.program.internal
...
 
+    override fun afterCollectionHandled(oCollection: Collection<ProgramIndicator>?) {
+        val inDbProgramIndicatorUids = programIndicatorStore.selectUids()
+        val apiProgramIndicatorUids = oCollection?.map(ProgramIndicator::uid)
+        val deleteProgramIndicatorUid = inDbProgramIndicatorUids.filter { inDbProgramIndicatorUid ->
+            val isPresentOnline = apiProgramIndicatorUids?.contains(inDbProgramIndicatorUid)
+            isPresentOnline == false
+        }
+        val query = WhereClauseBuilder().apply {
+            deleteProgramIndicatorUid.forEach { uid ->
```
vgarciabnz, line 59: There is a method `appendInKeyStringValues` that uses the IN operator and avoids the iteration of the array. But you should that `deleteProgramIndicatorUid` is not empty in first place

core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/APIDownloaderImpl.kt
```diff
@@ -74,6 +75,37 @@ internal class APIDownloaderImpl @Inject constructor(private val resourceHandler
...
+        )
+    }
+
+    private fun <P, O> downloadPartitionedWithoutHandling(
+        uids: Set<String>,
+        pageSize: Int,
+        pageDownloader: (Set<String>) -> Single<Payload<O>>,
+        transform: (O) -> P
+    ): Single<List<P>> {
+        val partitions = CollectionsHelper.setPartition(uids, pageSize)
```
vgarciabnz, line 97: You can reuse this method in the line 116 to avoid duplicated code

PR #1792, opened 2022-06-02
Update Debugging setup for Flipper

chaseTfreeman:
The dependencies are already included in the APK source code. Also I updated the path since there is a build.gradle in the root directory as well as one under /app. The build.gradle in root specified to NOT place the dependencies there.


chaseTfreeman:
Disregard this PR  as I was looking at the APK not the SDK

PR #1973, opened 2023-06-08
Androsdk 1689 network replace rx java in data set complete registration upload

andresmr:


core/src/androidTest/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationPostCallMockIntegrationShould.kt
```diff
@@ -0,0 +1,139 @@
...
+        dObject.blockingSet()
+        assertEquals(toPostRepository.blockingGet().size, 1)
+        assertEquals(syncedRepository.blockingGet().size, 0)
+
+        d2.dataSetModule().dataSetCompleteRegistrations().blockingUpload()
+
+        assertEquals(toPostRepository.blockingGet().size, 0)
+        assertEquals(syncedRepository.blockingGet().size, 1)
+
+        dhis2MockServer.enqueueMockResponse(200)
```
vgarciabnz, line 114: Checking the integration with v40, the response is 204 instead of 200. I am not sure if it was 200 before, or if it has been 204 for a while. It shouldn't change anything because it is a success code, but it is better to have it aligned and documented in the test
image

core/src/androidTest/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationPostCallMockIntegrationShould.kt
```diff
@@ -0,0 +1,139 @@
...
+    private fun provideDataSetCompleteRegistration() = DataSetCompleteRegistration.builder()
+        .period("2018")
+        .dataSet("BfMAe6Itzgt")
+        .attributeOptionCombo("bRowv6yZOF2")
+        .organisationUnit("DiszpKrYNg8")
+        .date(dateFormat.parse("2010-03-02"))
+        .storedBy("android")
+        .build()
+
+    private val dateFormat = SimpleDateFormat("yyyy-MM-dd")
```
vgarciabnz, line 138: Replace by `DateUtils.SIMPLE_DATE_FORMAT` to avoid adding Java dependencies. The implementation of SIMPLE_DATE_FORMAT still depends on Java, but it is centralized in a single class at least.

core/src/main/java/org/hisp/dhis/android/core/mockwebserver/Dhis2MockServer.java
```diff
@@ -418,8 +412,4 @@ public String getBaseEndpoint() {
     public RecordedRequest takeRequest() throws InterruptedException {
         return server.takeRequest();
     }
-
-    public void addResponse(String method, String path, String responseName, int responseCode) {
```
vgarciabnz, line 422: The methods `setDhis2Dispatcher()` and `addResponse()` are used in the Android Capture app for testing, they can't be removed.

andresmr: changed

PR #1985, opened 2023-07-11
fix: [ANDROSDK-1219] Escape ",", ";", ":" characters in queries' filters

dan-dhis2:
fix:
- ",", ";", ":" characters in filters need to be escaped with "/". The backend is fixed this issue only for the new importer, versions 40.1, 41.0 and newer.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelper.kt
@@ -160,7 +160,7 @@ internal class TrackedEntityInstanceQueryOnlineHelper @Inject constructor(
         fun toAPIFilterFormat(items: List<RepositoryScopeFilterItem>, upper: Boolean): List<String> {
             // Compatibility for the new Tracker (old Tracker will ignore this format)
             // Following characters need to be escaped escaped with a "/" for backend functionality
-            val charsToEscape  = "[;,:]".toRegex()
+            val charsToEscape = "[;,:]".toRegex()
 
             return items
                 .groupBy { it.key() }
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelperShould.kt
```diff
@@ -85,4 +86,33 @@ class TrackedEntityInstanceQueryOnlineHelperShould {
...
+    fun to_API_filter_format() {
+        // List of filters
+        val list = listOf(
+            "nom,app",
+            "nom:app",
+            "nom;app"
+        )
+
+        val expectedList = listOf(
+            "filterItem1:in:nom/,app/;nom/:app/;nom/;app",
```
vgarciabnz, line 100: For `in` operator, the expected string should be `nom/,app;nom/:app;nom/;app`. The char `;` separates the items in the list and must not be scaped.

core/src/test/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelperShould.kt
```diff
@@ -85,4 +86,33 @@ class TrackedEntityInstanceQueryOnlineHelperShould {
...
+            "filterItem2:like:nom/,app%%%<>%%%nom/:app%%%<>%%%nom/;app"
```
vgarciabnz, line 101: This string shows the internals of the method `listToStr`. It will fail if we change the separator in `FilterOperatorsHelper`, which should be independent to this test.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelper.kt
```diff
@@ -172,11 +175,13 @@ internal class TrackedEntityInstanceQueryOnlineHelper @Inject constructor(
...
         private fun getAPIValue(item: RepositoryScopeFilterItem): String {
+            val charsToEscape = "[;,:]".toRegex()
+
             return if (item.operator() == FilterItemOperator.IN) {
-                val list = FilterOperatorsHelper.strToList(item.value())
+                val list = FilterOperatorsHelper.strToList(item.value()).map { it.replace(charsToEscape, "/\$0") }
                 list.joinToString(";")
             } else {
-                item.value()
+                item.value().replace(charsToEscape, "/\$0")
```
vgarciabnz, line 184: The string `/\$0` is duplicated. It can be moved to a function such as `escapeAPIValue(value: String): String`.

PR #1986, opened 2023-07-11
refactor: [ANDROSDK-1677] network replace rx java in tracker download

andresmr:


core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
andresmr, line 1: I have surrounded `block.invoke()` with `runBlocking { }` to ensure that the `successfulTransaction` function is only called after all the flows in the block have completed.

core/src/main/java/org/hisp/dhis/android/core/tracker/exporter/TrackerDownloadCall.kt
```diff
@@ -50,88 +55,85 @@ import org.hisp.dhis.android.core.user.internal.UserOrganisationUnitLinkStore
...
-                    .switchMap {
-                        rxCallExecutor.wrapObservableTransactionally(
-                            Observable.merge(
-                                Observable.defer { downloadInternal(params, progressManager, relatives) },
-                                Observable.defer { downloadRelationships(progressManager, relatives) },
-                                Observable.fromCallable { progressManager.complete() }
-                            ),
-                            cleanForeignKeys = true
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun download(params: ProgramDataDownloadParams): Flow<TrackerD2Progress> = flow {
```
andresmr, line 65: Here is the issue, ` coroutineAPICallExecutor.wrapTransactionally` now is waiting to execute the block of flows (merged in a single flow) and then it finish the transaction, but it seems that the collect is not emiting, at least in the test in [EventDownloadCallMockIntegrationShould.kt](https://github.com/dhis2/dhis2-android-sdk/pull/1986/files#diff-66a45112c1eed316d562d8f4d713240f8c475bdebae91e0cb244f7bb14229412) is failing

core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
```diff
@@ -110,7 +110,9 @@ internal class CoroutineAPICallExecutorImpl @Inject constructor(
             val transaction = databaseAdapter.beginNewTransaction()
             try {
                 val result = coroutineScope {
-                    block.invoke()
+                    runBlocking {
```
andresmr, line 113: I have surrounded block.invoke() with runBlocking { } to ensure that the successfulTransaction function is only called after all the flows in the block have completed.

PR #1992, opened 2023-07-23
refactor: [ANDROSDK-1711] RxJava in ReservedValues call

dan-dhis2:
refactor:
- replace RxJava in ReservedValues call
- shorten test name "fill_up_100_values_if_db_does_not_have_at_least_50_when_no_number_of_values_to_reserve_is_specified" to pass ktlint restrictions


core/src/androidTest/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManagerRealIntegrationShould.kt
```diff
@@ -0,0 +1,354 @@
...
+    private var organisationUnit: OrganisationUnit? = null
+    private var pattern: String? = null
+
+    @Mock
+    var trackedEntityAttributeReservedValueQueryCallFactory: QueryCallFactory<TrackedEntityAttributeReservedValue,
+        TrackedEntityAttributeReservedValueQuery>? =
+        null
+
+    @Mock
+    var trackedEntityAttributeReservedValueCall: Callable<List<TrackedEntityAttributeReservedValue>>? =
```
vgarciabnz, line 80: We usually write this mock as `var trackedEntityAttributeReservedValueCall: Callable<List<TrackedEntityAttributeReservedValue>> = mock()`. In this way, it can be a `val` instead of a `var`.

core/src/androidTest/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManagerRealIntegrationShould.kt
```diff
@@ -0,0 +1,354 @@
...
+import org.mockito.Mockito
+
+@SuppressWarnings("MaxLineLength")
+class TrackedEntityAttributeReservedValueManagerRealIntegrationShould : BaseRealIntegrationTest() {
+    private var store: TrackedEntityAttributeReservedValueStore? = null
+    private val organisationUnitUid = "org_unit_uid"
+    private val programUid = "program_uid"
+    private val categoryComboUid = "category_combo_uid"
+    private val ownerUid = "xs8A6tQJY0s"
+    private var organisationUnit: OrganisationUnit? = null
```
vgarciabnz, line 71: It is initiated in setup, so it could be a nonNull lazy val.

vgarciabnz: Same for other properties in this block

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+) {
+    private val d2ProgressManager = D2ProgressManager(null)
+
+    /**
+     * @param attributeUid        Attribute uid
+     * @param organisationUnitUid Organisation unit uid
+     * @return Value of tracked entity attribute
+     * @see .getValue
+     */
+    suspend fun blockingGetValue(attributeUid: String, organisationUnitUid: String): String {
```
vgarciabnz, line 90: Public methods must not change the to `suspend` function because they are used by the app, sorry if it wasn't clear. The SDK must not change its public API at this level for 1.9.0.

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+    /**
+     * Get a reserved value and remove it from database. If the number of available values is below a threshold
+     * (default [.FILL_UP_TO] * [.FACTOR_TO_REFILL]) it tries to download before returning a value.
+     *
+     * @param attributeUid        Attribute uid
+     * @param organisationUnitUid Organisation unit uid
+     * @return Single with value of tracked entity attribute
+     */
+
+    suspend fun getValue(attributeUid: String, organisationUnitUid: String): String? {
```
vgarciabnz, line 103: This method should still return a `Single<String>` and remove the `suspend` modifier.

The same for other public methods, they must not change.

vgarciabnz (changes requested):

core/src/androidTest/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManagerRealIntegrationShould.kt
dan-dhis2: Let me check this one more time

core/src/androidTest/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManagerRealIntegrationShould.kt
dan-dhis2: Let me check this one more time

andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,409 @@
...
+     * Get a reserved value and remove it from database. If the number of available values is below a threshold
+     * (default [.FILL_UP_TO] * [.FACTOR_TO_REFILL]) it tries to download before returning a value.
+     *
+     * @param attributeUid        Attribute uid
+     * @param organisationUnitUid Organisation unit uid
+     * @return Single with value of tracked entity attribute
+     */
+
+    fun getValue(attributeUid: String, organisationUnitUid: String): Single<String> {
+        return rxSingle { getValueCoroutines(attributeUid, organisationUnitUid) ?: "" }
```
vgarciabnz, line 110: Previously, this method could return:
- an exception (if no more reserved values)
- the reserved value
- a null value (if for some reason there is a reservedValue with null value)

Now, the third case is defaulted to `""`, which changes the behavior and a sense that the value is correct, which is not right. It would be better to throw an exception or to return a null value. If we want to keep it the return value to strict NonNull, we can throw an exception instead

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
vgarciabnz: If `getValueCoroutines` returns NonNull, the expression `?: ""` is not needed anymore, is it?

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,407 @@
...
+        numberOfValuesToFillUp: Int?
+    ): Observable<D2Progress> {
+        return downloadReservedValuesFlow(attributeUid, numberOfValuesToFillUp).asObservable()
+    }
+
+    private fun downloadReservedValuesFlow(
+        attributeUid: String,
+        numberOfValuesToFillUp: Int?
+    ) = flow<D2Progress> {
+        downloadValuesForOrgUnits(attributeUid, numberOfValuesToFillUp).collect()
```
vgarciabnz, line 164: I think this flow does not emit anything. It should call `emit` or `emitAll`, something like 
`emitAll(downloadValuesForOrgUnits(attributeUid, numberOfValuesToFillUp))`

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,407 @@
...
+    /**
+     * Downloads reserved values for all the trackedEntityAttributeValues of type "generated", that is, it applies
+     * [.downloadReservedValues] for every generated attribute.
+     *
+     * @param numberOfValuesToFillUp An optional maximum number of values to reserve
+     * @return An Observable that notifies about the progress.
+     */
+
+    fun downloadAllReservedValues(numberOfValuesToFillUp: Int?): Observable<D2Progress> {
+        return rxObservable { downloadAllReservedValuesFlow(numberOfValuesToFillUp).collect() }
```
vgarciabnz, line 186: I think the use of `rxObservable` is not correct here. It emits with `send()`, something like 
`rxObservable { downloadAllReservedValuesFlow(numberOfValuesToFillUp).collect { send(it) } }`

It is easier to write it as:
`downloadAllReservedValuesFlow(numberOfValuesToFillUp).asObservable()`

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
dan-dhis2: I think these are just two ways of starting the flow, maybe `emitAll` looks more clean, but they should do the same

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
dan-dhis2: That was because of the `downloadAllReservedValuesFlow` design with a `return` value; I should've refactored it to just ```    private fun downloadAllReservedValuesFlow(
        numberOfValuesToFillUp: Int?
    ) = flow {
        val flows = generatedAttributes.map { attribute ->
            downloadValuesForOrgUnits(attribute.uid(), numberOfValuesToFillUp)
        }

        emitAll(flows.merge())
    }```

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    private suspend fun downloadValues(
+        trackedEntityAttributeUid: String,
+        organisationUnit: OrganisationUnit?,
+        numberToReserve: Int,
+        pattern: String?,
+        storeError: Boolean
+    ) {
+        return suspendCoroutine {
```
vgarciabnz, line 326: It seems that this `return` makes the call to never end. I think you can remove this whole line.

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+
+    /**
+     * Get the count of the reserved values by attribute. If a organisation unit uid is inserted as parameter the method
+     * will return the count of the reserved values by attribute and organisation unit.
+     *
+     * @param attributeUid        Attribute uid
+     * @param organisationUnitUid An optional organisation unit uid
+     * @return Single with the reserved value count by attribute or by attribute and organisation unit.
+     */
+    fun count(attributeUid: String, organisationUnitUid: String?): Int =
```
vgarciabnz, line 205: This method used to return a `Single<Int>` (check the Java docs)

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+        emitAll(downloadValuesForOrgUnits(attributeUid, numberOfValuesToFillUp))
+    }
+
+    /**
+     * @param numberOfValuesToFillUp An optional maximum number of values to reserve
+     * @see .downloadAllReservedValues
+     */
+    fun blockingDownloadAllReservedValues(numberOfValuesToFillUp: Int?) {
+        runBlocking {
+            downloadAllReservedValues(numberOfValuesToFillUp)
```
vgarciabnz, line 172: The code does not trigger the call. You can write either
`runBlocking { downloadAllReservedValuesFlow(numberOfValuesToFillUp).collect() }` (using the flow) or 
`downloadAllReservedValues(numberOfValuesToFillUp).blockingSubscribe()` (using the observable).

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -0,0 +1,406 @@
...
+    /**
+     * @param attributeUid           Attribute uid
+     * @param numberOfValuesToFillUp An optional maximum number of values to reserve
+     * @see .downloadReservedValues
+     */
+    fun blockingDownloadReservedValues(
+        attributeUid: String,
+        numberOfValuesToFillUp: Int?
+    ) {
+        runBlocking { downloadReservedValues(attributeUid, numberOfValuesToFillUp) }
```
vgarciabnz, line 134: Same issue than next comment https://github.com/dhis2/dhis2-android-sdk/pull/1992/files#diff-98a6a0146575ad8fe8fdbae85b6441ed5c6133152bb22b6ace80f939fb710d33R172

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2000, opened 2023-08-01
refactor: [ANDROSDK-1681] Refactor fileResourceDownloadCall to coroutines 

vgarciabnz:
Solves [ANDROSDK-1681](https://dhis2.atlassian.net/browse/ANDROSDK-1681)

### DIFF
--- a/core/src/test/java/org/hisp/dhis/android/core/tracker/importer/TrackerImporterFileResourcesPostCallShould.kt
+++ b/core/src/test/java/org/hisp/dhis/android/core/tracker/importer/TrackerImporterFileResourcesPostCallShould.kt
@@ -29,6 +29,8 @@ package org.hisp.dhis.android.core.tracker.importer
 
 import com.google.common.truth.Truth.assertThat
 import com.nhaarman.mockitokotlin2.*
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/hisp/dhis/android/core/tracker/importer/TrackerImporterFileResourcesPostCallShould.kt
```diff
@@ -49,16 +51,9 @@ class TrackerImporterFileResourcesPostCallShould {
...
-        fileResourcePostCall = TrackerImporterFileResourcesPostCall(
-            fileResourcesPostCall,
-            fileResourceHelper
-        )
-    }
-
+    @OptIn(ExperimentalCoroutinesApi::class)
     @Test
-    fun `Should create a single post call for repeated attributes`() {
+    fun `Should create a single post call for repeated attributes`() = runTest {
```
vgarciabnz, line 56: The `runTest` method makes possible to mock suspend functions using the same syntax than before. There is no need to use the `stub` and `onBlocking` methods.

PR #2017, opened 2023-09-08
refactor: [ANDROSDK-1732] replace RxJava for ConstantModule

dan-dhis2:
[ANDROSDK-1732]
refactor:
- ConstantModule calls from RxJava to Coroutines
fix:
- AuthorityEndpointCallFactory to access Kotlin properties
style:
- Added line to Eof

[ANDROSDK-1732]: https://dhis2.atlassian.net/browse/ANDROSDK-1732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/constant/internal/ConstantService.kt
```diff
@@ -25,21 +25,19 @@
...
+internal interface ConstantService {
     @GET("constants")
-    Call<Payload<Constant>> constants(@Query("fields") @Which Fields<Constant> fields,
-                                      @Query("paging") boolean paging);
-}
\ No newline at end of file
+    suspend fun constants(
+        @Query("fields") @Which fields: Fields<Constant>,
+        @Query("paging") paging: Boolean
+    ): Payload<Constant>?
```
vgarciabnz, line 42: Could it be NonNull?

core/src/main/java/org/hisp/dhis/android/core/arch/modules/internal/TypedModuleDownloader.kt
```diff
@@ -0,0 +1,34 @@
...
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.arch.modules.internal
+
+import io.reactivex.Single
+
+interface TypedModuleDownloader<O> {
+    fun downloadMetadata(): Single<O>?
```
vgarciabnz, line 33: Could it return a NonNull object?

core/src/main/java/org/hisp/dhis/android/core/arch/call/factories/internal/ListCoroutineCallFactoryImpl.kt
```diff
@@ -0,0 +1,47 @@
...
+    protected val data: GenericCallData,
+    protected val coroutineAPICallExecutor: CoroutineAPICallExecutor
+) : ListCoroutineCallFactory<P> {
+    override suspend fun create(): List<P> {
+        val objects: List<P> = fetcher().invoke()
+        processor().process(objects)
+        return objects
+    }
+
+    protected abstract fun fetcher(): suspend () -> List<P>
```
vgarciabnz, line 45: Maybe it could written as `protected abstract suspend fun fetcher(): List<P>`, so the method `invoke()` could be skipped in the previous method and it would be simpler.

core/src/main/java/org/hisp/dhis/android/core/constant/internal/ConstantModuleDownloader.kt
```diff
@@ -25,30 +25,23 @@
...
-    }
-
-    @Override
-    public Single<List<Constant>> downloadMetadata() {
-        return Single.fromCallable(constantCallFactory.create());
+class ConstantModuleDownloader @Inject internal constructor(
+    private val constantCallFactory: ConstantCoroutineCallFactory
+) :
+    TypedModuleDownloader<List<Constant>> {
+    override fun downloadMetadata(): Single<List<Constant>> {
```
vgarciabnz, line 42: It seems this class could be marked as `internal`, and this method could return a suspend function instead of a `Single`. The class `MetadataCall` is ready to accept suspend functions in their methods.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/constant/internal/ConstantModuleDownloader.kt
dan-dhis2: Yes it could be internal, but for using coroutines, I think we either have to remove `TypeModuleDownloader` or refactor it to Coroutines (it's used by a couple more classes)

vgarciabnz: It could be duplicated if we don't want to migrate all the usages of `TypeModuleDownloader`

dan-dhis2: I'd refactor it to use coroutines, so we can have one more thing migrated

core/src/main/java/org/hisp/dhis/android/core/domain/metadata/MetadataCall.kt
```diff
@@ -130,15 +130,15 @@ internal class MetadataCall @Inject constructor(
         useCaseDownloader.downloadMetadata().blockingAwait()
         emit(progressManager.increaseProgress(StockUseCase::class.java, false))
 
-        constantModuleDownloader.downloadMetadata().blockingGet()
+        runBlocking { constantModuleDownloader.downloadMetadata() }
```
vgarciabnz, line 133: The `runBlocking` is not needed. This line is inside a flow, so you can call suspend functions directly.

core/src/main/java/org/hisp/dhis/android/core/domain/metadata/MetadataCall.kt
```diff
@@ -130,15 +130,15 @@ internal class MetadataCall @Inject constructor(
...
         emit(progressManager.increaseProgress(Constant::class.java, false))
 
         smsModule.configCase().refreshMetadataIdsCallable().blockingAwait()
         emit(progressManager.increaseProgress(SmsModule::class.java, false))
     }
 
     private fun executeUserCallAndChildren(progressManager: D2ProgressManager): Flow<D2Progress> = flow {
-        val user = userModuleDownloader.downloadMetadata().blockingGet()
+        val user = runBlocking { userModuleDownloader.downloadMetadata() }
```
vgarciabnz, line 141: The same here

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2018, opened 2023-09-11
refactor: [ANDROSDK-1736] reserved values from RxJava to Coroutines

dan-dhis2:
refactor:
- All RxJava calls to Coroutines for ReservedValues
fix:
- EndpointCall.kt extending


core/src/main/java/org/hisp/dhis/android/core/arch/call/fetchers/internal/ListNoResourceWithErrorCatcherCallFetcher.kt
```diff
@@ -25,30 +25,20 @@
...
-    }
-
-    protected abstract retrofit2.Call<List<P>> getCall();
-
-    @Override
-    public final List<P> fetch() throws D2Error {
-        return apiCallExecutor.executeObjectCallWithErrorCatcher(getCall(), errorCatcher);
+    @Throws(D2Error::class)
+    suspend fun fetch(): List<P> {
+        return coroutineAPICallExecutor.wrap(errorCatcher = errorCatcher) { call }.getOrThrow()
```
vgarciabnz, line 42: This `call` isn't a function, it is a val, I am not sure it would work fine. Anyway, this class is never used, it can be removed.

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueService.kt
```diff
@@ -25,28 +25,28 @@
...
+import retrofit2.http.Path
+import retrofit2.http.Query
 
+interface TrackedEntityAttributeReservedValueService {
     @GET("trackedEntityAttributes/{" + TRACKED_ENTITY_ATTRIBUTE_UID + "}/generateAndReserve")
-    Call<List<TrackedEntityAttributeReservedValue>> generateAndReserve(
-            @Path(TRACKED_ENTITY_ATTRIBUTE_UID) String trackedEntityAttributeUid,
-            @Query("numberToReserve") Integer numberToReserve);
+    suspend fun generateAndReserve(
+        @Path(TRACKED_ENTITY_ATTRIBUTE_UID) trackedEntityAttributeUid: String?,
```
vgarciabnz, line 38: All these parameters could be NonNull

core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttributeReservedValueManager.kt
```diff
@@ -311,7 +308,7 @@ class TrackedEntityAttributeReservedValueManager @Inject internal constructor(
         if (remainingValues < minNumberToTryFill) {
             val numberToReserve = fillUpTo!! - remainingValues
             downloadValues(
-                attribute, organisationUnit, numberToReserve, pattern, storeError
```
vgarciabnz, line 314: The purpose of this storeError was to optionally store the error based on the method that originated the query: if the query is generated as a side effect of `getValue()`, it is fine if the query fails because the device can be offline and the user didn't trigger the query. If the query is generated explicitly by the user, it is important to store the error.

The `storeError` property can be added to the TrackedEntittyAttributeReservedValueQuery and then applied in the coroutineExecutor.

vgarciabnz (changes requested):
This PR has conflicting files

core/src/main/java/org/hisp/dhis/android/core/arch/call/factories/internal/QueryCoroutineCallFactoryImpl.kt
```diff
@@ -34,9 +34,9 @@ import org.hisp.dhis.android.core.arch.call.queries.internal.BaseQueryKt
 
 internal abstract class QueryCoroutineCallFactoryImpl<P, Q : BaseQueryKt> protected constructor(
     protected val data: GenericCallData,
-    protected val coroutineAPICallExecutor: CoroutineAPICallExecutor
+    protected val coroutineAPICallExecutor: CoroutineAPICallExecutor,
 ) : QueryCallFactory<P, Q> {
-    override suspend fun create(query: Q): MutableList<P> {
+    override suspend fun create(query: Q): List<P> {
         val objects: List<P> = fetcher(query)
```
vgarciabnz, line 40: Just for convention, it should be `fetcher(query).fetch()` instead of calling the `fetch()` method inside the `fetcher()`.

core/src/main/java/org/hisp/dhis/android/core/arch/call/factories/internal/QueryCoroutineCallFactoryImpl.kt
```diff
@@ -34,9 +34,9 @@ import org.hisp.dhis.android.core.arch.call.queries.internal.BaseQueryKt
...
         processor(query).process(objects)
         return objects.toMutableList()
```
vgarciabnz, line 42: This `toMutableList()` is not needed, isn't it?

vgarciabnz (changes requested):

dan-dhis2:
Closing this PRs due to messed git history

PR #2019, opened 2023-09-11
refactor: [ANDROSDK-1720] refactor Call classes to Kotlin

dan-dhis2:
Refactor the following 
- UidsCall
- QueryCall
- ListCall
- ObjectCall
And their implementations, to Kotlin


dan-dhis2:
Closed due to messed git history

PR #2022, opened 2023-09-14
(experimental) refactor: [ANDROSDK-1736] reserved values from RxJava to Coroutines 

dan-dhis2:
Marked as _experimental_ as it comes from [PR 2018](https://github.com/dhis2/dhis2-android-sdk/pull/2018), which had some messed git history parts

refactor:
- All RxJava calls to Coroutines for ReservedValues
fix:
- EndpointCall.kt extending


core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueEndpointCallFactory.kt
```diff
@@ -0,0 +1,84 @@
...
+                return if (query.organisationUnit() == null) {
+                    service.generateAndReserve(
+                        query.trackedEntityAttributeUid(),
+                        query.numberToReserve(),
+                    )
+                } else {
+                    service.generateAndReserveWithOrgUnitCode(
+                        query.trackedEntityAttributeUid(),
+                        query.numberToReserve(),
+                        query.organisationUnit()!!.code()!!,
```
vgarciabnz, line 68: The property `code()` might be null. We can check it in the "if" condition.
In java, the code won't fail with NullPointerException and the parameter will just be ignored because it is null. Now, it would throw a NullPointer

core/src/main/java/org/hisp/dhis/android/core/arch/call/factories/internal/QueryCoroutineCallFactoryImpl.kt
```diff
@@ -25,34 +25,23 @@
...
+        val objects: List<P> = fetcher(query)
+        processor(query).process(objects)
+        return objects
     }
 
-    protected abstract CallFetcher<P> fetcher(Q query);
-    protected abstract CallProcessor<P> processor(Q query);
-}
\ No newline at end of file
+    protected abstract suspend fun fetcher(query: Q): List<P>
```
vgarciabnz, line 45: Following the convention, this would be the `CoroutineCallFetcher` you created in the other PR. In the `create()` method, it will call to `fetcher(query).fetch()`.

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2033, opened 2023-09-29
refactor: [ANDROSDK-1717] Refactor collection repositories to kotlin - Part 3

marcamsn:
Solves [ANDROSDK-1717](https://dhis2.atlassian.net/browse/ANDROSDK-1717).

[ANDROSDK-1717]: https://dhis2.atlassian.net/browse/ANDROSDK-1717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/event/search/EventQueryCollectionRepository.kt
```diff
@@ -0,0 +1,315 @@
...
+
+    override fun getPaged(pageSize: Int): LiveData<PagedList<Event>> {
+        return eventCollectionRepository.getPaged(pageSize)
+    }
+
+    override fun getPagingData(pageSize: Int): Flow<PagingData<Event>> {
+        return eventCollectionRepository.getPagingData(pageSize)
+    }
+
+    fun getDataSource(): DataSource<Event, Event> {
```
vgarciabnz, line 288: This property should be something like:
```
val dataSource: DataSource<Event, Event>
        get() = eventCollectionRepository.dataSource
```
This is just to avoid breaking changes in the API

core/src/main/java/org/hisp/dhis/android/core/event/search/EventQueryCollectionRepository.kt
```diff
@@ -0,0 +1,315 @@
...
+    private val connectorFactory: ScopedFilterConnectorFactory<
+        EventQueryCollectionRepository,
+        EventQueryRepositoryScope,
+        >
+    private val eventFilterRepository: EventFilterCollectionRepository
+
+    @JvmField
+    val scope: EventQueryRepositoryScope
+
+    init {
```
vgarciabnz, line 75: I think this "init" block can be removed. Properties "eventFilterRepository" and "scope" can be used directly from the constructor by adding "private val". The "connectorFactory" can be initialized inline with the declaration

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2043, opened 2023-10-10
refactor: [ANDROSDK-1756] ProgramModuleDownloader remove RxJava

dan-dhis2:
refactor:
- Remove RxJava for ProgramModuleDownloader
- All related calls which were tied to these changes


core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramModuleDownloader.kt
```diff
@@ -61,53 +61,48 @@ internal class ProgramModuleDownloader @Inject constructor(
...
+            ProgramParentUidsHelper
+                .getAssignedTrackedEntityAttributeUids(programs, trackedEntityTypes),
+        )
+
+
+        val optionSetUids = ProgramParentUidsHelper.getAssignedOptionSetUids(
+            attributes,
+            programStages,
+        )
+        flowOf(
```
vgarciabnz, line 88: I think this `flowOf` doesn't add any value, does it? I know it follows the same structure than before, but I don't see the point in keeping this calls together. They don't run in parallel in any case, I we don't want them to run in parallel because of the database lock

core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramModuleDownloader.kt
```diff
@@ -61,53 +61,48 @@ internal class ProgramModuleDownloader @Inject constructor(
...
-                                ).ignoreElements()
-                            }
-                            .concatWith(downloadFiltersAndWorkingLists(programUids))
-                    }
-                }
-        }
     }
 
-    private fun downloadFiltersAndWorkingLists(programUids: Set<String>): Completable {
-        return Single.merge(
```
vgarciabnz, line 105: The same here. The point on using `Single.merge` was that it avoids the concatenation with `flatMap`. In this way, the list of calls is flatten. 
Using coroutines, this is not needed anymore because the call can be written in sequence.

core/src/test/java/org/hisp/dhis/android/core/program/internal/ProgramEndpointCallShould.kt
```diff
@@ -0,0 +1,117 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(JUnit4::class)
+class ProgramEndpointCallShould : BaseCallShould() {
+
+    private val programService: ProgramService = mock()
+    private val programHandler: ProgramHandler = mock()
+
+    @Captor
+    private val fieldsCaptor: ArgumentCaptor<Fields<Program>>? = null
```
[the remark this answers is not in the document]
dan-dhis2: I reversed the test, as it was not necessary for now to refactor it. Thank you!

core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramStageService.kt
```diff
@@ -25,21 +25,21 @@
...
+internal fun interface ProgramStageService {
     @GET("programStages")
-    Single<Payload<ProgramStage>> getProgramStages(@Query("fields") @Which Fields<ProgramStage> fields,
-                                                   @Query("filter") String programUidsFilterString,
-                                                   @Query("filter") String accessDataReadFilter,
-                                                   @Query("paging") Boolean paging);
-}
\ No newline at end of file
+    suspend fun getProgramStages(
+        @Query("fields") @Which fields: Fields<ProgramStage?>?,
```
vgarciabnz, line 40: All these properties could be non null

core/src/main/java/org/hisp/dhis/android/core/validation/internal/ValidationRuleUidsCallImpl.kt
```diff
@@ -42,29 +40,25 @@ internal class ValidationRuleUidsCallImpl @Inject constructor(
...
+    override suspend fun download(uids: Set<String>): List<ObjectWithUid> {
+        return uids.flatMap {
+            apiDownloader.downloadLink(
+                it,
+                linkHandler,
+                {
+                    service.getDataSetValidationRuleUids(
+                        it,
+                        BaseIdentifiableObject.UID,
+                        Boolean.FALSE,
```
vgarciabnz, line 54: This Boolean.FALSE could be just false

core/src/main/java/org/hisp/dhis/android/core/program/internal/ProgramCall.kt
```diff
@@ -58,6 +58,11 @@ internal class ProgramCall @Inject internal constructor(
         }
     }
 
+    // TODO : remove this fun and refactor ProgramEndpointCallShould to Kotlin
+    fun coroutineEncapsulationForJavaTest(uids: Set<String>): List<Program> = runBlocking {
```
vgarciabnz: This method could be removed

core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/APIDownloaderImpl.kt
```diff
@@ -88,85 +87,84 @@ internal class APIDownloaderImpl @Inject constructor(private val resourceHandler
...
+        partitions.forEach { partition ->
+            val transformedItems = pageDownloader(partition).items().map { transform(it) }
+            results.addAll(transformedItems)
+        }
+
+        return results
     }
 
-    private fun <P, O> downloadPartitionedWithCustomHandling(
+    // TODO : remove the Single making sure it is not needed by the callers
```
vgarciabnz: This comment, does it still make sense?

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2047, opened 2023-10-18
feat: [ANDROSDK-1699] Replace Dagger by Koin

vgarciabnz:
Solves [ANDROSDK-1699](https://dhis2.atlassian.net/browse/ANDROSDK-1699)

[ANDROSDK-1699]: https://dhis2.atlassian.net/browse/ANDROSDK-1699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz (commented):
The general approach was to use Koin annotations using the `@Singleton` keyword. Koin accepts both `@Singleton` and `@Single`, they are synonyms. I preferred `@Singleton` because `@Single` clashes with the RxJava class.

There are big changes in androidTests because the general structure doesn't fit well with Koin because it is not possible to instantiate the Koin app several times. For this reason, there is a single D2 object and the different scenarios are handled as different account (different server urls).

dan-dhis2 (approved):
I'd add the comments to the Dependency injection modules

PR #2049, opened 2023-10-19
refactor: [ANDROSDK-1731-BASE-METHODS] Replace RxJava in pending tracker methods

vgarciabnz:
Refactor some pending methods in Tracker download call.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/event/internal/EventPersistenceCallFactory.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/event/internal/EventPersistenceCallFactory.kt
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.android.core.event.internal
 
-import io.reactivex.Completable
 import org.hisp.dhis.android.core.arch.handlers.internal.IdentifiableDataHandlerParams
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/hisp/dhis/android/core/event/internal/EventPersistenceCallFactory.kt
```diff
@@ -41,36 +40,32 @@ internal class EventPersistenceCallFactory(
...
-            }
+    ) {
+        val params = IdentifiableDataHandlerParams(
+            hasAllAttributes = false,
+            overwrite = false,
+            asRelationship = asRelationship,
+        )
+        eventHandler.handleMany(events, params, relatives)
+        if (hasMissingOrganisationUnitUids(events)) {
+            organisationUnitDownloader.refreshOrganisationUnits().blockingAwait()
```
vgarciabnz, line 63: Once OrganisationUnit module is refactored to coroutines, you will have to remove this `blockingAwait()` in two places: in the class and in the equivalent for TEIs. @dan-dhis2

dan-dhis2 (approved):

dan-dhis2 (approved):

PR #2055, opened 2023-10-25
refactor: [ANDROSDK-1754] ConfigCase from Java/RxJava to Kotlin/Coroutines

dan-dhis2:
refactor:
- ConfigCase.java to ConfigCase.kt
- Remove RxJava from ConfigCase main calls


core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
```diff
@@ -0,0 +1,191 @@
...
+     * Method to download metadata ids. This is required before using the SMS module.
+     */
+    suspend fun refreshMetadataIds() {
+
+        val enabled = localDbRepository.isModuleEnabled.await()
+
+        if (enabled) {
+            try {
+                val config = metadataDownloadConfig.await()
+                val metadata = webApiRepository.getMetadataIds(config).await()
```
vgarciabnz, line 149: This is the API call, which is still a Single. We should get down to API level. The idea behind this refactor is to remove RxJava dependencies at API level.

core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
```diff
@@ -0,0 +1,191 @@
...
+        /**
+         * Get the metadata download configuration.
+         * @return `Single with the` metadata download configuration.
+         */
+        get() = localDbRepository.metadataDownloadConfig
+
+    /**
+     * Method to download metadata ids. This is required before using the SMS module.
+     */
+    suspend fun refreshMetadataIds() {
```
vgarciabnz, line 142: This is part of the public API, which should not be modified. It should return a `Completable`.

core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
```diff
@@ -0,0 +1,191 @@
...
+            Log.d(TAG, "Not refreshing SMS metadata, because sms module is disabled")
+        }
+
+    }
+
+    /**
+     * Callable that triggers the method [.refreshMetadataIds].
+     * @return Callable object to refresh metadata ids.
+     */
+    suspend fun refreshMetadataIdsCallable() {
```
vgarciabnz, line 165: This is part of the internal API, we can use `suspend` here and modify the method with `internal`.

core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
```diff
@@ -0,0 +1,191 @@
...
+     * @param metadataIdsConfig Configuration with the metadata ids.
+     * @return `Completable` that completes when the configuration is changed.
+     */
+    fun setMetadataDownloadConfig(metadataIdsConfig: GetMetadataIdsConfig?): Completable {
+        return if (metadataIdsConfig == null) {
+            Completable.error(IllegalArgumentException("Received null config"))
+        } else localDbRepository.setMetadataDownloadConfig(metadataIdsConfig)
+    }
+
+    val metadataDownloadConfig: Single<GetMetadataIdsConfig>
```
vgarciabnz, line 132: By default, functions starting with `fun getX()` are translated to `val x`, which follows Java conventions but is a bit weird... What do you think? Should we stick to `fun getMetadataDownloadConfig()`?

core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.hisp.dhis.android.core.sms.domain.repository.internal.LocalDbRepository
+
+/**
+ * Used to set initial data that is common for all sms sending tasks
+ */
+class ConfigCase(
+    private val webApiRepository: WebApiRepository,
+    private val localDbRepository: LocalDbRepository
+) {
+    val smsModuleConfig: Single<SmsConfig>
```
vgarciabnz, line 45: Same comment as below for `getMetadataDownloadConfig`

vgarciabnz (changes requested):
ConfigCase is part of the public API, so their methods' signature should remain the same. The only internal method is `refreshMetadataIdsCallable()`, which is used internally.

core/src/main/java/org/hisp/dhis/android/core/sms/domain/interactor/ConfigCase.kt
dan-dhis2: Agree, the `fun` approach seems more solid and clear

dan-dhis2:
This PR has been reverted to **Draft**, as it still needs analysis and some problems arose with `rxCompletable`

vgarciabnz (approved):

PR #2128, opened 2024-03-20
fix: [ANDROSDK-1837] TrackerLineList fails to evaluate filtered lastUpdated

vgarciabnz:
Solves [ANDROSDK-1837](https://dhis2.atlassian.net/browse/ANDROSDK-1837)

[ANDROSDK-1837]: https://dhis2.atlassian.net/browse/ANDROSDK-1837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/evaluator/NotSupportedEvaluator.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/evaluator/NotSupportedEvaluator.kt
@@ -31,7 +31,7 @@ package org.hisp.dhis.android.core.analytics.trackerlinelist.internal.evaluator
 internal class NotSupportedEvaluator : TrackerLineListEvaluator() {
 
     override fun getCommonSelectSQL(): String {
-        return "Not supported"
+        return "'Not supported'"
     }
 
     override fun getCommonWhereSQL(): String {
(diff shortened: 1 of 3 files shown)


marcamsn (approved):

core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/evaluator/NotSupportedEvaluator.kt
```diff
@@ -31,7 +31,7 @@ package org.hisp.dhis.android.core.analytics.trackerlinelist.internal.evaluator
 internal class NotSupportedEvaluator : TrackerLineListEvaluator() {
 
     override fun getCommonSelectSQL(): String {
-        return "Not supported"
+        return "'Not supported'"
```
andresmr, line 34: why do we need the comas?

andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/evaluator/NotSupportedEvaluator.kt
vgarciabnz: The return value is directly used to build a SQL query. If it is not wrapped with commas, it is considered as a column or an operator. It is necessary to wrap it with commas to be considered a string

PR #2134, opened 2024-04-18
[ANDROSDK-1852] add root certificates

andresmr:
Add Let's encrypt root certificates and add it programmatically for versions under API 24


vgarciabnz:
The CI was failing because of clearText communication with the mock server. It is solved by adding "cleartextTraffic = true" in the network configuration, although it is not the ideal solution because it opt-in for clearText by default, which might not be the desired behavior by same apps. 
Before this change, the SDK didn't specify any value for "cleartextTraffic" and it was the android app who specifies a value in the Manifest > application section. This is ideal because it lets the app decide whether cleartextTraffic should be allowed. 
If we can't find a way to make it work without explicitly enabling cleartext, we can go with this. The Android app enables cleartextTraffiic, so for this app it won't be a change. If we wo with this, we should include it in the release notes.

PR #2148, opened 2024-05-15
refactor: [ANDROSDK-1855] Refactor WhereClauseBuilder class

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/evaluator/OrganisationUnitEvaluator.kt
```diff
@@ -94,10 +94,10 @@ internal class OrganisationUnitEvaluator(
             }
 
             is OrganisationUnitFilter.Level -> {
-                val level = orgunitLevelStore.selectByUid(filter.uid)
+                val level = orgunitLevelStore.selectByUid(filter.uid)!!
```
andresmr, line 97: could this nonNull asserting lead to NullPointerException?

vgarciabnz: If the level does not exist, it should fail somehow. We could use a specific exception for this parameter that is already defined:
```
val level = orgunitLevelStore.selectByUid(filter.uid)
                 ?: throw AnalyticsException.InvalidOrganisationUnitLevel(filter.uid)
```

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2149, opened 2024-05-20
refactor: [ANDROSDK-1856] Refactor SQLStatementBuilder classes

taridepaco:


vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/db/querybuilders/internal/SQLStatementBuilderImpl.kt
```diff
@@ -0,0 +1,222 @@
...
+import org.hisp.dhis.android.core.arch.db.sqlorder.internal.SQLOrderType
+import org.hisp.dhis.android.core.arch.db.stores.projections.internal.LinkTableChildProjection
+import org.hisp.dhis.android.core.arch.db.tableinfos.TableInfo
+import org.hisp.dhis.android.core.arch.helpers.CollectionsHelper
+import org.hisp.dhis.android.core.common.CoreColumns
+import org.hisp.dhis.android.core.common.IdentifiableColumns
+
+// TODO save TableInfo instead of separate files when architecture 1.0 is ready
+@Suppress("SpreadOperator", "TooManyFunctions")
+class SQLStatementBuilderImpl internal constructor(
```
vgarciabnz, line 39: This class should be internal. Currently, only the primary constructor method is internal.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/db/querybuilders/internal/SQLStatementBuilder.kt
```diff
@@ -25,38 +25,29 @@
...
-    String deleteWhere();
+package org.hisp.dhis.android.core.arch.db.querybuilders.internal
+
+import org.hisp.dhis.android.core.arch.db.stores.projections.internal.LinkTableChildProjection
+
+@Suppress("TooManyFunctions")
+internal interface SQLStatementBuilder : ReadOnlySQLStatementBuilder {
+    fun getTableName(): String
+    fun getColumns(): Array<String>
+    fun selectUids(): String?
```
vgarciabnz, line 36: Sorry, I didn't notice it before. This should return a non-null String. It is better to change it now to avoid cascading nulability errors when we migrate other parts of the code

vgarciabnz (changes requested):

vgarciabnz (approved):

marcamsn (approved):

PR #2151, opened 2024-05-22
refactor: [ANDROSDK-1857] Refactor Filter classes in API package

taridepaco:


core/src/test/java/org/hisp/dhis/android/core/arch/api/filters/internal/FilterConverterShould.java
```diff
@@ -141,58 +141,6 @@ public void returns_correct_path_when_create_a_retrofit_request_using_filters_an
...
-    public void returns_correct_path_when_create_a_retrofit_request_ignoring_null_filter() throws IOException, InterruptedException {
-        ArrayList<String> values = new ArrayList<>(2);
-        values.add("uid1");
-        values.add("uid2");
-
-        TestService service = testService(server);
-
-        service.test(
-                InFilter.create(Field.create("id"), values),
-                SingleValueFilter.gt(Field.create("lastUpdated"), null)
```
vgarciabnz, line 154: I think this test still makes sense. This second property could be null to prove that a null Filter is ignored in the generated url.

core/src/test/java/org/hisp/dhis/android/core/arch/api/filters/internal/FilterConverterShould.java
```diff
@@ -141,58 +141,6 @@ public void returns_correct_path_when_create_a_retrofit_request_using_filters_an
...
-    public void returns_correct_path_when_create_a_retrofit_request_ignoring_empty_string_filter() throws IOException, InterruptedException {
-        ArrayList<String> values = new ArrayList<>(2);
-        values.add("uid1");
-        values.add("uid2");
-
-        TestService service = testService(server);
-
-        service.test(
-                InFilter.create(Field.create("id"), values),
-                SingleValueFilter.gt(Field.create("lastUpdated"), "")
```
vgarciabnz, line 172: Maybe we have to give it a second thought because an empty string in a SingleValueFilter will lead to an invalid url. For example, this request throws an Internal Server Error `api/dataElements?filter=id:eq:&fields=id,name`.

core/src/main/java/org/hisp/dhis/android/core/arch/api/filters/internal/SingleValueFilter.kt
```diff
@@ -25,48 +25,39 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
+internal class SingleValueFilter<T, K> (
+    override val field: Field<T, K>,
+    override val operator: String,
+    override val values: Collection<String>,
+) : Filter<T, K> {
+    override fun generateString(): String {
+        return "${field.name()}:$operator:${values.first()}"
```
vgarciabnz, line 38: This method will always have a collection with a single value. But the constructor is internal so it is possible that other parts of the code create instances of this class with an empty list. Could it be possible to make the constructor internal?

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2155, opened 2024-05-28
refactor: [ANDROSDK-1858] Refactor Field classes

taridepaco:


core/src/test/java/org/hisp/dhis/android/core/arch/api/fields/internal/FieldShould.kt
```diff
@@ -0,0 +1,101 @@
...
+class FieldShould {
+//    @Test(expected = NullPointerException::class)
+//    fun throw_null_pointer_exception_on_null_field_name() {
+//        Field.create<Any, Any>(null)
+//    }
+//
+//    @Test(expected = NullPointerException::class)
+//    fun throw_null_pointer_exception_on_null_nested_field_name() {
+//        NestedField.create<Any, Any>(null)
+//    }
```
vgarciabnz, line 47: These tests, why are they commented out? If they don't make sense any longer, we can remove them

core/src/main/java/org/hisp/dhis/android/core/arch/api/fields/internal/FieldsConverter.java
```diff
@@ -46,45 +44,8 @@ public String convert(Fields fields) throws IOException {
 
         // recursive function which processes
         // properties and builds query string
-        append(builder, (List<Property>) fields.fields());
+        Fields.Companion.append(builder, fields.getFields());
```
vgarciabnz, line 47: Ideally we could have a single line with something such as `fields.generateString()` (in the same way we have for FilterConverter)

core/src/main/java/org/hisp/dhis/android/core/arch/api/fields/internal/Fields.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    companion object {
+        @JvmStatic
+        fun <K> builder(): Builder<K> {
+            return Builder()
+        }
+
+        @Suppress("NestedBlockDepth")
+        fun append(builder: StringBuilder, properties: List<Property<*, *>>) {
```
vgarciabnz, line 60: This method could be simplified a lot with Kotlin. Anyway, it can be done a separate PR if you want

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/api/fields/internal/Fields.kt
```diff
@@ -25,58 +25,50 @@
...
-            return this;
+        fun build(): Fields<T> {
+            return Fields(fields.toList())
         }
+    }
 
-        public final Builder<T> fields(@NonNull Collection<Property<T, ?>> properties) {
-            if (properties == null || properties.isEmpty()) {
-                throw new IllegalArgumentException("properties null or empty collection");
+    fun generateString(properties: List<Property<*, *>>?): String {
```
vgarciabnz, line 53: It is a bit weird to have to call it as `.generateString(null)` from the outside. The parameter is only needed for the recursion, isn't it? 

I can see two possible approaches to avoid it:
- Make the parameter optional, such as `generateString(properties: List<Property<*,*>>? = null)`. In this way, it might be called as `.generateString()`.
- Create a private method in the companion object with this logic, maybe called `private fun generateStringFromFields()` and then call it from here as `generateStringFromFields(fields)`.

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2157, opened 2024-06-05
feat: [ANDROSDK-1691] Download FileResource information in a single query

taridepaco:


core/src/sharedTest/resources/trackedentity/tracked_entity_attribute_value_image_resources.json
```diff
@@ -0,0 +1,43 @@
+{
+  "FileResource": [
```
vgarciabnz, line 2: The name of the property is `fileResources`. It is not so important because the deserializer takes whichever property is present in the payload containing an array of FileResources. Anyway, it is better to keep it as it is in the api.

Check https://play.im.dhis2.org/stable-2-41-0/api/fileResources?filter=id:in:[KWM8rwyXlt3,TBge3Ikhdiq]&paging=false&fields=id,name,created,lastUpdated,storageStatus,domain

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2162, opened 2024-06-12
refactor: [ANDROSDK-1863] Migrate module interfaces to Kotlin

taridepaco:


core/src/androidTest/java/org/hisp/dhis/android/core/settings/SettingsModuleMockIntegrationShould.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.hisp.dhis.android.core.utils.integration.mock.BaseMockIntegrationTestFullDispatcher
+import org.hisp.dhis.android.core.utils.runner.D2JunitRunner
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(D2JunitRunner::class)
+class SettingsModuleMockIntegrationShould : BaseMockIntegrationTestFullDispatcher() {
+    @Test
+    fun allow_access_to_system_setting() {
+        var systemSettings: List<SystemSetting?> = d2.settingModule().systemSetting().blockingGet()
```
marcamsn, line 41: Use `var` only for variables that will be modified later; otherwise, it's better to use `val`.

core/src/androidTest/java/org/hisp/dhis/android/testapp/sms/SmsModuleMockIntegrationShould.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.hisp.dhis.android.core.utils.integration.mock.BaseMockIntegrationTestEmptyEnqueable
+import org.hisp.dhis.android.core.utils.runner.D2JunitRunner
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(D2JunitRunner::class)
+class SmsModuleMockIntegrationShould : BaseMockIntegrationTestEmptyEnqueable() {
+    @Test
+    fun access_submit_case() {
+        var submitCase = d2.smsModule().smsSubmitCase()
```
marcamsn, line 40: Maybe it could be done in one line without losing readability, right?

core/src/main/java/org/hisp/dhis/android/core/category/CategoryModule.kt
```diff
@@ -0,0 +1,37 @@
...
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.category
+
+interface CategoryModule {
+    fun categories(): CategoryCollectionRepository
+    fun categoryOptions(): CategoryOptionCollectionRepository
+    fun categoryOptionCombos(): CategoryOptionComboCollectionRepository
+    fun categoryCombos(): CategoryComboCollectionRepository
+
```
marcamsn, line 35: Remove line break for consistency

core/src/main/java/org/hisp/dhis/android/core/settings/SettingModule.kt
```diff
@@ -0,0 +1,50 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.settings
+
+interface SettingModule {
+    fun systemSetting(): SystemSettingCollectionRepository
+
```
marcamsn, line 32: Remove the line break for consistency

taridepaco:
@vgarciabnz Be careful, two unused classes were removed (MetadataModuleDownloader and MetadataModuleByUidDownloader ) and two public module interfaces, that were inside an internal folder, were moved to their immediate parent folder (ImportModule and WipeModule).

vgarciabnz:
> two public module interfaces, that were inside an internal folder, were moved to their immediate parent folder (ImportModule and WipeModule).

About this, the change in the package of the public interfaces could cause breaking changes in the apps. It is not very likely to happen because those modules are usually directly consumed (`d2.wipeModule().wipeData()`) and not stored in a variable, but still I would make this change in a minor version at least. Maybe it is change for 1.11.

vgarciabnz (approved):

PR #2166, opened 2024-06-17
chore: [ANDROSDK-1864] Remove not useful child type parameter from fields

marcamsn:
Solves [ANDROSDK-1864](https://dhis2.atlassian.net/browse/ANDROSDK-1864).

[ANDROSDK-1864]: https://dhis2.atlassian.net/browse/ANDROSDK-1864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/arch/api/fields/internal/FieldsConverterFactory.kt
```diff
@@ -25,32 +25,17 @@
...
-            }
-        }
-        return null;
-    }
+internal class FieldsConverterFactory : Converter.Factory() {
+    override fun stringConverter(
+        type: Type,
+        annotations: Array<Annotation>,
+        retrofit: Retrofit,
+    ): Converter<*, String>? = annotations.firstOrNull { it is Which }?.let { FieldsConverter() }
```
taridepaco, line 40: Now only the first annotation is checked for type Which, but previously the array of annotations was looped on until one was of type Which. Is this intended?

marcamsn: Hey! `firstOrNull` usually helps to safely retrieve the first element of a collection or sequence, or return null if the collection is empty. But you can also use `firstOrNull` with a predicate to find the first element that matches a given condition. If no such element exists, it returns null.

In this case the condition is `it is Which`, so it loops the annotations until it finds that annotation.

vgarciabnz (approved):

PR #2170, opened 2024-06-25
chore: [ANDROSDK-1875] Refactor Http Exception and Response to reduce retrofit dependencies

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/user/internal/UserAccountDisabledErrorCatcher.kt
```diff
@@ -58,29 +57,18 @@ internal class UserAccountDisabledErrorCatcher(
...
-            val responseErrorBody = objectMapper.readValue(errorBody, HttpMessageResponse::class.java)
+            val isUnauthorized = response.statusCode == HttpURLConnection.HTTP_UNAUTHORIZED
+            val responseErrorBody = objectMapper.readValue(response.errorBody, HttpMessageResponse::class.java)
             isUnauthorized && responseErrorBody.message().contains("Account disabled")
         } catch (e: Exception) {
             false
         }
     }
 
     fun catchError(throwable: Throwable): D2ErrorCode? {
```
vgarciabnz, line 70: This method was a bit confusing to me because this casting from `Throwable` to `D2HttpException` is not obvious. Looking at the code it seems that this method (and the one below) are not used. As this class is internal, we can remove them to clean up the code

vgarciabnz (approved):

PR #2172, opened 2024-06-27
chore: [ANDROSDK-1877]: Add Ktor  and migrate ConstantService from Retrofit

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/mockwebserver/Dhis2MockServer.java
```diff
@@ -443,8 +432,4 @@ public String getBaseEndpoint() {
     public RecordedRequest takeRequest() throws InterruptedException {
         return server.takeRequest();
     }
-
-    public void addResponse(String method, String path, String responseName, int responseCode) {
```
vgarciabnz, line 447: Do not remove this method, it is used in the app to mock responses. This class and this method are intentionally public.

core/src/main/java/org/hisp/dhis/android/core/constant/internal/KtorConstantService.kt
```diff
@@ -0,0 +1,52 @@
...
+        val url = "${URL}constants"
+        val parameters = mapOf(
+            "fields" to fields.generateString(),
+            "paging" to paging.toString(),
+        )
+        return client.get(url, parameters)
+    }
+
+    companion object {
+        private const val URL = "https://temporary-dhis-url.org/api/"
```
vgarciabnz, line 50: I think this variable could be included in the KtorServiceClient to avoid repetition.

core/src/main/java/org/hisp/dhis/android/core/constant/internal/KtorConstantService.kt
```diff
@@ -0,0 +1,52 @@
...
+@Singleton
+internal class KtorConstantService(private val client: KtorServiceClient) {
+    suspend fun constants(fields: Fields<Constant>, paging: Boolean): Payload<Constant> {
+        val url = "${URL}constants"
+        val parameters = mapOf(
+            "fields" to fields.generateString(),
+            "paging" to paging.toString(),
+        )
+        return client.get(url, parameters)
+    }
```
vgarciabnz, line 47: Maybe not for now, but I think we can avoid repetition of parameter names all across the services. For example, the parameters `fields` or `paging` are used in almost every method.

Just brainstorming, we could use a DSL to call the method something like:
```
return client.get {
  url("constants")
  parameters {
    fields(fields)
    paging(paging)
  }
}
```
Then, the method in KtorServiceClient will translate those parameters to `fields` and `paging`. And it will call the method `.generateString()` in the fields.

We can talk about that in another moment.

vgarciabnz (changes requested):
Great PR! The only required change is the method in the mockServer because it affects the app. The others are just comments.

core/src/main/java/org/hisp/dhis/android/core/mockwebserver/Dhis2MockServer.java
taridepaco: And what about the other methods removed in this script?

vgarciabnz: The other methods are private, if they are not used the can be removed

vgarciabnz (approved):

PR #2186, opened 2024-07-19
chore: [ANDROSDK-1891] Expose an alternative method for API interaction

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/D2.kt
```diff
@@ -74,8 +74,7 @@ import org.hisp.dhis.android.core.wipe.internal.WipeModule
 class D2 internal constructor(internal val d2DIComponent: D2DIComponent) {
     private val modules: D2Modules = d2DIComponent.modules
 
-    @VisibleForTesting
-    internal fun ktor(): KtorServiceClient {
+    fun ktor(): HttpServiceClient {
```
vgarciabnz, line 77: This method is exposed and is named `ktor`, shouldn't it be called `httpServiceClient`?

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/HttpServiceClient.kt
```diff
@@ -44,11 +44,11 @@ import okhttp3.ResponseBody.Companion.toResponseBody
 import org.koin.core.annotation.Singleton
 
 @Singleton
-internal class KtorServiceClient(
-    private val client: HttpClient,
+class HttpServiceClient(
+    val client: HttpClient,
```
vgarciabnz, line 48: These two properties are now exposed in the public API. The `client` property is ktor-specific. Couldn't they be hidden by marking them as internal or private?

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/HttpServiceClient.kt
```diff
@@ -44,11 +44,11 @@ import okhttp3.ResponseBody.Companion.toResponseBody
...
     var baseUrl: String = "https://temporary-dhis-url.org/api/",
 ) {
-    internal suspend inline fun <reified T> request(
+    suspend inline fun <reified T> request(
```
vgarciabnz, line 51: If this method is public, the parameter `requestMethod` is exposed, which exposes a ktor class `HttpMethod`. Couldn't it be hidden? Maybe the annotation PublishedApi can help here

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/HttpServiceClient.kt
taridepaco: It seems Public-API inline function cannot access non-public-API internal variables.

vgarciabnz: the publishedAPI annotation doesn't help here?

andresmr (approved):

vgarciabnz (approved):

PR #2195, opened 2024-07-31
feat: [ANDROSDK-1615-3] Remove period generators based on Calendar

vgarciabnz:
This PR solves [ANDROSDK-1615](https://dhis2.atlassian.net/browse/ANDROSDK-1615).

It removes the old period generator classes and uses the new ones in their place.

[ANDROSDK-1615]: https://dhis2.atlassian.net/browse/ANDROSDK-1615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/period/clock/internal/ClockProviderFactory.kt
```diff
@@ -25,31 +25,26 @@
...
-    protected void moveToStartOfCurrentPeriod() {
-        // do nothing
+    fun setRegular() {
+        clockProvider = RegularClockProvider()
     }
 
-    @Override
-    protected void moveToStartOfCurrentYear() {
-        calendar.set(Calendar.DAY_OF_YEAR, 1);
+    fun setFixed() {
```
taridepaco, line 41: This function and the one below seem to be devoted for testing only. Couldn't they be moved to a test helper? The fixed date below seems a bit arbitrary if left here. Also, the original one sets 11 for the month instead of 12, but it probably doesn't matter.

vgarciabnz: The original month set to 11 actually was December (zero-based). But the day was 1-based. It is one of the tricky things about Calendar.

Makes sense to move it to a test helper, I will do it

taridepaco (approved):

PR #2197, opened 2024-08-05
chore: [ANDROSDK-1901] Remove java utils from PeriodParser

marcamsn:
Removes `Calendar`, `Date`, `Matcher` and `Pattern` from `PeriodParser`.
Solves [ANDROSDK-1901](https://dhis2.atlassian.net/browse/ANDROSDK-1901).

[ANDROSDK-1901]: https://dhis2.atlassian.net/browse/ANDROSDK-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/arch/d2/internal/JavaDIClasses.kt
```diff
@@ -47,7 +47,7 @@ internal val javaDIClasses = module {
     single { FileResourceCleaner(get(), get(), get()) }
     single { NoteUniquenessManager(get()) }
     single { PeriodHelper(get(), get(), get(), get(), get()) }
-    single { PeriodParser(get()) }
+    single { PeriodParser() }
```
vgarciabnz, line 50: This entry could be removed if the class PeriodParser is annotated with @Singleton. It is not possible to annotate Java classes, that is why it was here.
Anyway, it can be replaced in separate task to refactor all these classes to kotlin

vgarciabnz (approved):

PR #2198, opened 2024-08-05
chore: [ANDROSDK-1894] Move Okhttp interceptors to ktor plugins

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/PreventURLDecodePlugin.kt
```diff
@@ -25,29 +25,30 @@
...
 
-import okhttp3.Interceptor
-import okhttp3.Request
-import okhttp3.Response
-import java.io.IOException
+import io.ktor.client.plugins.api.createClientPlugin
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.http.takeFrom
+
+object PreventURLDecodePlugin {
```
vgarciabnz, line 35: It might be internal

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/ServerURLVersionRedirectionPlugin.kt
```diff
@@ -0,0 +1,79 @@
...
+import io.ktor.client.request.HttpRequest
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.client.request.takeFrom
+import io.ktor.client.statement.HttpResponse
+import io.ktor.http.HttpHeaders
+import org.hisp.dhis.android.core.arch.api.internal.DynamicServerURLPlugin.transformRequest
+import org.hisp.dhis.android.core.arch.api.internal.HttpStatusCodes.REDIRECT_MAX
+import org.hisp.dhis.android.core.arch.api.internal.HttpStatusCodes.REDIRECT_MIN
+
+object ServerURLVersionRedirectionPlugin {
```
vgarciabnz, line 43: It might be internal

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/UserAgentPlugin.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.os.Build
+import io.ktor.client.*
+import io.ktor.client.plugins.*
+import io.ktor.client.request.*
+import io.ktor.http.HttpHeaders
+import io.ktor.util.*
+import org.hisp.dhis.android.BuildConfig
+import org.hisp.dhis.android.core.D2Configuration
+
+public class UserAgentPlugin private constructor(public val agent: String) {
```
vgarciabnz, line 38: Could it be internal?

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/HttpServiceClientFactory.kt
```diff
@@ -34,11 +34,15 @@ import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.http.ContentType
 import io.ktor.serialization.jackson.JacksonConverter
 import okhttp3.OkHttpClient
+import org.hisp.dhis.android.core.D2Configuration
+import org.hisp.dhis.android.core.arch.api.authentication.internal.ParentAuthenticatorPlugin
 import org.hisp.dhis.android.core.arch.json.internal.ObjectMapperFactory
 
-object KtorFactory {
-    fun ktor(
+object HttpServiceClientFactory {
```
vgarciabnz, line 41: It might be internal

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/DynamicServerURLPlugin.kt
```diff
@@ -25,37 +25,35 @@
...
+import io.ktor.client.plugins.api.createClientPlugin
+import io.ktor.client.request.*
+import io.ktor.http.*
+import io.ktor.util.*
 
-    companion object {
-        fun transformRequest(request: Request): Request {
-            val transformedUrl = transformUrl(request.url.toString())?.let { it.toHttpUrlOrNull() }
-            return transformedUrl?.let { request.newBuilder().url(it).build() } ?: request
+object DynamicServerURLPlugin {
```
vgarciabnz, line 37: It might be internal

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/KtorPlugins.kt
```diff
@@ -25,33 +25,29 @@
...
+internal fun HttpClientConfig<*>.addKtorPlugins(
+    d2Configuration: D2Configuration,
+    authenticator: ParentAuthenticatorPlugin,
+) {
+    install(UserAgentPlugin) { this.d2Configuration = d2Configuration }
+    install(DynamicServerURLPlugin.instance)
+    install(ServerURLVersionRedirectionPlugin.instance)
+    install(PreventURLDecodePlugin.instance)
+    install(authenticator.instance)
+    install(HttpTimeout) {
```
vgarciabnz, line 48: I am not sure if read/write timeouts are exactly equivalent. They might not have a direct equivalent in Ktor.

Maybe we can leave the timeout configuration in OkHttp because the D2Configuration object will be different for Android and iOS. In Ktor, we could apply the maximum value of them, just to be sure we don't set a lower timeout.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/KtorPlugins.kt
taridepaco: That was the best equivalent I could find, but I agree that we can keep it as it was.

vgarciabnz (approved):

PR #2200, opened 2024-08-07
feat: [ANDROSDK-1897] Create new properties for map configuration 

taridepaco:
Related task [ANDROSDK-1897](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/93?selectedIssue=ANDROSDK-1897)

[ANDROSDK-1897]: https://dhis2.atlassian.net/browse/ANDROSDK-1897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/androidTest/java/org/hisp/dhis/android/testapp/settings/AppearanceSettingsObjectRepositoryMockIntegrationShould.kt
```diff
@@ -139,6 +139,34 @@ class AppearanceSettingsObjectRepositoryMockIntegrationShould :
         assertThat(program2Setting?.itemHeader()?.programIndicator()).isNull()
     }
 
+    @Test
+    fun should_return_minimumLocationAccuracy_settings() {
+        val setting = d2.settingModule().appearanceSettings().getGlobalProgramConfigurationSetting()
+        assertThat(setting?.minimumLocationAccuracy()).isEqualTo(2)
```
andresmr, line 145: I think that minimum location accuracy it's 5

core/src/main/java/org/hisp/dhis/android/core/settings/ProgramConfigurationSetting.java
```diff
@@ -92,6 +98,10 @@ public abstract static class Builder {
 
         public abstract Builder itemHeader(ProgramItemHeader itemHeader);
 
+        public abstract Builder minimumLocationAccuracy(Integer minimumLocationAccuracy);
+
+        public abstract Builder disableManualLocation(Boolean disableManualLocation);
```
andresmr, line 103: Does this property a default value or is it nullable? Should it has a default value?

marcamsn: I think it should be enabled by default.

vgarciabnz: Most of the properties here are nullable and are not given a default value in the SDK. They have a default initial value in the ASWA (the initial config) and the application gives them a default value when they are null. I would keep the same behavior for this property. 
We could consider that the SDK should be the responsible for setting the default value, but if we do so, it should be done for all the properties. So I would keep this property nullable for now

vgarciabnz (approved):

PR #2208, opened 2024-08-21
chore: [ANDROSDK-1900] Remove calendar usages from TrackedEntityAttributeReservedValueValidatorHelper

marcamsn:
Solves [ANDROSDK-1900](https://dhis2.atlassian.net/browse/ANDROSDK-1900).

[ANDROSDK-1900]: https://dhis2.atlassian.net/browse/ANDROSDK-1900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueValidatorHelper.kt
```diff
@@ -0,0 +1,88 @@
...
+package org.hisp.dhis.android.core.trackedentity.internal
+
+import kotlinx.datetime.*
+import org.hisp.dhis.android.core.period.clock.internal.ClockProviderFactory
+import org.koin.core.annotation.Singleton
+
+@Singleton
+internal class TrackedEntityAttributeReservedValueValidatorHelper {
+    @Throws(IllegalStateException::class)
+    fun getExpiryDateCode(pattern: String?): Instant {
```
vgarciabnz, line 37: This method is quite hard to understand. It was already hard before, but it had named parameters at least. It is not clear what first, second and third mean.

Maybe naming the parameters in `nextExpiryDate(yearly = flags.first,...)` would help a little bit.

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueValidatorHelper.kt
```diff
@@ -0,0 +1,88 @@
...
+            monthly -> {
+                val nextMonth = now.plus(1, DateTimeUnit.MONTH)
+                LocalDate(nextMonth.year, nextMonth.month, 1)
+            }
+
+            yearly -> {
+                LocalDate(now.year + 1, 1, 1)
+            }
+
+            else -> error("No expiry date available for this pattern.")
```
vgarciabnz, line 80: I didn't know about these method and didn't understand the IllegalStateException at the header. They seemed to be handy shortcuts to throw some exceptions, we can start using them more often (https://kotlinlang.org/docs/exceptions.html#throw-exceptions-with-precondition-functions)

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueValidatorHelper.kt
marcamsn: I've made some changes to this file in the PR I'm currently working on, so how about we merge this PR and I'll add the parameters to the new one?

vgarciabnz (approved):

PR #2210, opened 2024-08-22
feat: [ANDROSDK-1908] Add property displayContent in programRuleAction

taridepaco:
Related task: [ANDROSDK-1908](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/93?selectedIssue=ANDROSDK-1908)

[ANDROSDK-1908]: https://dhis2.atlassian.net/browse/ANDROSDK-1908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andresmr (approved):

vgarciabnz (changes requested):
There is a missing change in a test `ProgramRuleActionShould`. It is already implicitly tested in the integration test, but still good to have it as a separate unit test.

vgarciabnz:
This PR adds a new property, which makes the app change from `content` to `displayContent` @andresmr . The alternative would be deprecate `content` and fallback to `displayContent` value, as we did for the labels in program and programStage. What do you think is the best approach? Keeping both properties is more verbose but maybe more consistent with the rest of the code.

vgarciabnz (approved):

taridepaco:
This property was added in version 2.33.

PR #2218, opened 2024-09-03
chore: [ANDROSDK-1913] migrate arch.call package to kotlin. Part 2

taridepaco:
Related task: [ANDROSDK-1913](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/93?selectedIssue=ANDROSDK-1913)

[ANDROSDK-1913]: https://dhis2.atlassian.net/browse/ANDROSDK-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/event/internal/EventQuery.kt
```diff
@@ -25,19 +25,18 @@
...
-    public abstract Set<String> uids();
-
-    public static UidsQuery create(Set<String> uids) {
-        return new AutoValue_UidsQuery(1, BaseQuery.DEFAULT_PAGE_SIZE, false, uids);
-    }
-}
+internal data class EventQuery(
+    val commonParams: TrackerQueryCommonParams,
+    val orgUnit: String? = null,
+    val uids: Collection<String?> = emptyList(),
```
vgarciabnz, line 36: Could it be `Collection<String>` (nonNull)?

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueEndpointCallFactory.kt
```diff
@@ -72,8 +72,8 @@ internal class TrackedEntityAttributeReservedValueEndpointCallFactory internal c
         return TrackedEntityAttributeReservedValueCallProcessor(
             data.databaseAdapter,
             handler,
-            query.organisationUnit(),
-            query.trackedEntityAttributePattern(),
+            query.organisationUnit,
+            query.trackedEntityAttributePattern!!,
```
vgarciabnz, line 76: This `!!` is risky. The parameter in the query could be nonNull. Or check the nullability in a previous method.

Maybe it was already like this before and this parameter is never null, but it still seems a potential issue.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/event/internal/EventQuery.kt
taridepaco: Yes, in fact this internal class is only used in its own test. Maybe it could be removed?

vgarciabnz: Really? If so, it could be removed

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityAttributeReservedValueEndpointCallFactory.kt
taridepaco: Downstream, the code is prepared to receive a nullable pattern, so I'll remove the not-null assertion operator and allow the pattern to be null there.

vgarciabnz (approved):

PR #2222, opened 2024-09-06
chore: [ANDROSDK-1914] Remove uses of D2CallExecutor

taridepaco:
Related task: [ANDROSDK-1914](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/93?selectedIssue=ANDROSDK-1914)

[ANDROSDK-1914]: https://dhis2.atlassian.net/browse/ANDROSDK-1914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz:
This PR removes the usage of Java Callable, which is the whole point of the JIRA ticket and that is enough. Additionally, do you think it makes sense to remove the `D2CallExecutor` class? Or is it better to keep a blocking alternative to the `CoroutineCallExecutor`?

taridepaco:
I found a way of not having to make the functions "suspend" just by using a lambda function instead of the Callable. So the change is quite simple and doesn't affect many other scripts. It can be kept for now, in my opinion.

vgarciabnz (approved):

PR #2225, opened 2024-09-27
fix: [ANDROSDK-1923] Replace Ktor plugins by Okhttp interceptors 

vgarciabnz:
Solves [](https://dhis2.atlassian.net/browse/ANDROSDK-1923)

This PR address two different issue that made the SDK not compatible with Android API < 26:
- Usage of @Deprecation.since() method. Removed and replaced by a comment.
- Usage of ktor-serialization-jackson-jvm version 2.3.11, which depends on Jackson 2.16.1. It has been downgraded to Ktor 2.1.3, which depends on Jackson 2.13.3 (we were using 2.13.4 before).

The use of Ktor 2.1.3 forced us to remove the plugins (the new plugin API was introduced in 2.2.0). The workaround was to use the OkHttp interceptors again.


vgarciabnz:
Pending to implement the functionality related to absoluteUrls.

vgarciabnz:
Verify the commented-out test LogInOfflineCallMockIntegrationShould @taridepaco

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/PreventURLDecodeInterceptor.java
```diff
@@ -26,29 +26,31 @@
...
+public class PreventURLDecodeInterceptor implements Interceptor {
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        String encodedUrl = request.url().toString();
 
-    private fun replaceEncodedCharacters(request: HttpRequestBuilder) {
-        val encodedUrl = request.url.toString()
+        String nonEncodedUrl = encodedUrl
```
andresmr, line 44: Maybe is not valuable, but in order to escalate replacements you can do a map of replacements and then a fold:

`val replacements = mapOf(
    "%2C" to ",",
    "%5B" to "[",
    "%5D" to "]",
    "%3A" to ":"
)`

so you can add more replacements in the future, and then:
`val nonEncodedUrl = replacements.entries.fold(encodedUrl) { acc, (encoded, decoded) ->
    acc.replace(encoded, decoded)
}`

also you could create an extension function

core/src/test/java/org/hisp/dhis/android/core/arch/api/httpservice/internal/RequestBuilderShould.kt
```diff
@@ -75,7 +75,8 @@ class RequestBuilderShould {
     fun build_url_correctly() = runTest {
         val mockEngine = MockEngine { request ->
             assertEquals("https://temporary-dhis-url.org/api/test", request.url.toString())
-            assertThat(request.attributes.contains(isAbsouteUrlAttributeKey)).isFalse()
+//            assertThat(request.attributes.contains(isAbsouteUrlAttributeKey)).isFalse()
```
andresmr, line 80: please remove commented code

taridepaco (approved):

PR #2226, opened 2024-10-03
fix: [ANDROSDK-1924] Update lastest version algorithm to match logic in the APK Distribution app

taridepaco:
Related task: [ANDROSDK-1924](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/93?selectedIssue=ANDROSDK-1924)

[ANDROSDK-1924]: https://dhis2.atlassian.net/browse/ANDROSDK-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/settings/internal/LatestAppVersionCall.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/settings/internal/LatestAppVersionCall.kt
@@ -46,22 +46,27 @@ internal class LatestAppVersionCall(
 
     override suspend fun tryFetch(storeError: Boolean): Result<LatestAppVersion, D2Error> {
         return coroutineAPICallExecutor.wrap(storeError = storeError) {
-            val userGroupUids = userModule.userGroups().blockingGetUids()
-
-            val versions = settingAppService.versions().items()
-
-            val filteredVersions = versions.filter { version ->
-                version.userGroups?.any { userGroupUid ->
-                    userGroupUids.contains(userGroupUid)
-                } ?: false
-            }
-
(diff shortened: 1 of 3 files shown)


andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/settings/internal/LatestAppVersionCall.kt
```diff
@@ -46,22 +46,27 @@ internal class LatestAppVersionCall(
...
 
-            version?.let { LatestAppVersion.builder().version(it.version).downloadURL(it.downloadURL).build() }
-                ?: settingAppService.latestAppVersion()
+        val filteredVersions = versions.filter { version ->
+            version.userGroups?.any { userGroupUid ->
+                userGroupUids.contains(userGroupUid)
+            } ?: false
         }
+
+        val availableVersions = filteredVersions + versions.find { it.isDefault == true }!!
```
vgarciabnz, line 67: The configuration should include a version with the property `isDefault` set to true. If there is no default version, this line will throw a NPE because of the `!!`.

I will try to make the method more tolerant to unexpected configurations. For example:
- if there is no default version but the user has any group with an assigned version, use that version.
- if there are more than one default versions, take the highest one (instead of the first one).

If we wanted to validate that there is just one default version, that is valid too. In that case I will throw an explicit `InvalidArgumentException` with a message instead of a generic NPE caused by `!!`.

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2240, opened 2024-11-11
fix: [ANDROSDK-1922] improve image downloading behavior considering content lenght

taridepaco:
Related task: [ANDROSDK-1922](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-1922)

[ANDROSDK-1922]: https://dhis2.atlassian.net/browse/ANDROSDK-1922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileResizerHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileResizerHelper.kt
@@ -141,15 +141,10 @@ object FileResizerHelper {
         LARGE(1024),
     }
 
-    open class DimensionSizeB(val name: String, val maxSizeB: Long) {
-        companion object {
-            val SMALL = FixedImageOption("SMALL", 400000L)
-            val MEDIUM = FixedImageOption("MEDIUM", 1600000L)
-            const val ORIGINAL_NAME = "ORIGINAL"
-
-            fun create(name: String, maxSizeB: Long): DimensionSizeB = CustomImageOption(name, maxSizeB)
-        }
-        class FixedImageOption(name: String, maxSizeB: Long) : DimensionSizeB(name, maxSizeB)
-        private class CustomImageOption(name: String, maxSizeB: Long) : DimensionSizeB(name, maxSizeB)
+    @Suppress("MagicNumber")
+    internal sealed class DimensionSizeB(val name: String, val maxSizeB: Long) {
+        object Small : DimensionSizeB("SMALL", 400000L)
+        object Medium : DimensionSizeB("MEDIUM", 1600000L)
+        data class Original(val originalMaxSizeB: Long) : DimensionSizeB("ORIGINAL", originalMaxSizeB)
     }
 }
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileResizerHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileResizerHelper.kt
@@ -142,9 +142,14 @@ object FileResizerHelper {
     }
 
     @Suppress("MagicNumber")
-    internal sealed class DimensionSizeB(val name: String, val maxSizeB: Long) {
-        object Small : DimensionSizeB("SMALL", 400000L)
-        object Medium : DimensionSizeB("MEDIUM", 1600000L)
-        data class Original(val originalMaxSizeB: Long) : DimensionSizeB("ORIGINAL", originalMaxSizeB)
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCall.kt
```diff
@@ -277,16 +280,17 @@ internal class FileResourceDownloadCall(
...
-        download: suspend (V) -> ByteArray?,
+        download: suspend (V, Long?) -> ByteArray?,
         fileResource: FileResource,
     ): FileResource? {
+        val contentLength = fileResource.contentLength()
         val acceptedContentLength = (maxContentLength == null) ||
-            (fileResource.contentLength() == null) ||
-            (fileResource.contentLength()!! <= maxContentLength)
+            (contentLength == null) ||
+            (min(fileResource.contentLength()!!.toLong(), DimensionSizeB.SMALL.maxSizeB) <= maxContentLength)
```
andresmr, line 289: could you avoid using !! changing `fileresource.contentLenght()` by `contentLenght`?

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCall.kt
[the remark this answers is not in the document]
taridepaco: True, I thought about the non image types at some point, but it went out of my head.

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceUtil.kt
```diff
@@ -146,4 +149,38 @@ internal object FileResourceUtil {
...
+        // Filter out sizes requiring upscaling
+        val validSizes = sizes.filter { it.maxSizeB <= orignalContentLength }
+
+        // Step 3: Remove ORIGINAL if it exceeds MEDIUM max size
+        val limitedSizes = validSizes.filterNot {
+            it.name == originalSize.name && it.maxSizeB > DimensionSizeB.Medium.maxSizeB
+        }
+
+        // Select the largest size within maxContentLength
+        return limitedSizes.lastOrNull { it.maxSizeB <= maxContentLength }?.name ?: limitedSizes.first().name
```
vgarciabnz, line 173: The `first()` method might return a `NoSuchElementException`when the list is empty. This will never happen because the application will only reach this part of the code when the contentLength is accepted, which implicitly implies that this list won't be empty. But it seems a bit fragile to me.

Another option could be to change the condition and use this method to determine if the `contentLength` is accepted or not. We could make a difference between FILE and IMAGE (which didn't exist before this PR):
- if it is a FILE, return ORIGINAL only if the contentLength is smaller than maxContentLength. Otherwise return null.
- if it is an IMAGE, return the corresponding dimension, having MEDIUM as the maximum. If there is no valid dimension, return null.

Then, this dimension could be used to determine if the contentLength is accepted: if the dimension is not null, it is accepted. Otherwise, it is not. From this point on, we will have a nonNull dimension and we could get rid of this.

Additionally, if the dimension is ORIGINAL we should skip the parameter `dimension` in the api request. We were always including this parameter (even for files) and it only makes sense for images.

andresmr (approved):

vgarciabnz (approved):

core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileResizerHelper.kt
```diff
@@ -140,4 +140,16 @@ object FileResizerHelper {
...
+
+    @Suppress("MagicNumber")
+    internal sealed class DimensionSize(val name: String, val maxSizeB: Long) {
+        object Small : DimensionSize("SMALL", 400000L)
+        object Medium : DimensionSize("MEDIUM", 1600000L)
+        object NotSupported : DimensionSize("NOT_SUPPORTED", 0L)
+        data class Original(val originalMaxSizeB: Long) : DimensionSize(ORIGIANL_NAME, originalMaxSizeB)
+
+        companion object {
+            const val ORIGIANL_NAME = "ORIGINAL"
```
vgarciabnz, line 152: I noticed this typo but not important at all. I've just approved it

PR #2253, opened 2024-12-03
chore: [ANDROSDK-1947] Create boilerplate for network layer and migrate two models

taridepaco:
Related task: [ANDROSDK-1947](https://dhis2.atlassian.net/browse/ANDROSDK-1947)

[ANDROSDK-1947]: https://dhis2.atlassian.net/browse/ANDROSDK-1947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz (changes requested):
Good PR!

I have added a few comments. Apart from that, the checks have failed because of the API validator (first time we see it in action). It complaints about the new classes in the `network` package, they are public and are added to the public API. That is a good catch, I think all of them can be internal (in general, everything under `network` could probably be internal)

core/src/main/java/org/hisp/dhis/android/core/constant/internal/ConstantCoroutineCallFactory.kt
```diff
@@ -40,14 +40,14 @@ import org.koin.core.annotation.Singleton
 internal class ConstantCoroutineCallFactory(
     data: GenericCallData,
     coroutineAPICallExecutor: CoroutineAPICallExecutor,
-    private val service: ConstantService,
     private val handler: ConstantHandler,
+    private val constnatGetCall: ConstantGetCallInterface,
```
marcamsn, line 44: Hey! There's a typo on `constnatGetCall` . 😄

marcamsn: Maybe `ConstantProvider`, `ConstantFetcher`, `ConstantCallProvider`, `ConstantCallGetter` are better names. 
Also maybe the val property could be `callProvider`,  `callFetcher` or `callGetter`.

marcamsn: Also, we try to avoid use the `Interface` suffix and add a `Impl` suffix on the implementation when there is only one.

core/src/main/java/org/hisp/dhis/android/network/option/OptionDTO.kt
```diff
@@ -0,0 +1,57 @@
...
+
+@Serializable
+internal data class OptionDTO(
+    @SerialName("id") override val uid: String,
+    override val code: String? = null,
+    override val name: String? = null,
+    override val displayName: String? = null,
+    override val created: String = "",
+    override val lastUpdated: String = "",
+    override val deleted: Boolean? = null,
```
marcamsn, line 53: Can all this overrides belong to the interface?

internal interface BaseIdentifiableObjectInterface {
    val uid: String
    val code: String? get() = null
    val name: String? get() = null
    val displayName: String? get() = null
    val created: String? get() = ""
    val lastUpdated: String? get() = ""
    val deleted: Boolean? get() = false
}

core/src/main/java/org/hisp/dhis/android/network/constant/ConstantGetCall.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.hisp.dhis.android.core.constant.internal.ConstantGetCallInterface
+import org.hisp.dhis.android.network.common.HttpServiceClientKotlinx
+import org.koin.core.annotation.Singleton
+
+@Singleton
+internal class ConstantGetCall(
+    private val httpClient: HttpServiceClientKotlinx,
+    private val service: ConstantService = ConstantService(httpClient),
+) : ConstantGetCallInterface {
+    override suspend fun getCall(): List<Constant> {
```
marcamsn, line 42: Instead of `getCall` maybe name the method as `getConstants` so we can have later a `postConstants`, `deleteConstants` and others like this.

core/src/main/java/org/hisp/dhis/android/network/constant/ConstantDTO.kt
```diff
@@ -0,0 +1,47 @@
...
+internal class ConstantPayload(
+    override val pager: Pager? = null,
+    @SerialName("constants") override val items: List<ConstantDTO> = emptyList(),
+) : Payload<ConstantDTO>(pager, items)
+
+@Serializable
+internal data class ConstantDTO(
+    @SerialName("id") override val uid: String,
+    val value: Double? = null,
+) : BaseIdentifiableObjectDTO()
```
vgarciabnz, line 47: So, is it possible to inherit from BaseIdentifiableObjectDto? Does it need to be Serializable as well?

core/src/main/java/org/hisp/dhis/android/core/arch/api/payload/internal/PayloadInterface.kt
```diff
@@ -0,0 +1,40 @@
...
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.api.payload.internal
+
+internal interface PayloadInterface<T> {
```
vgarciabnz, line 31: Is it possible to have this interface called Payload? And the other two would be `JacksonPayload` and  `JsonPayload` or something like that.

core/src/main/java/org/hisp/dhis/android/network/constant/ConstantDTO.kt
taridepaco: It's possible but it requires the annotation, otherwise the parent fields are not used for deserializing

vgarciabnz (approved):

PR #2261, opened 2024-12-16
chore: [ANDROSDK-1949-PART-2] Refactor should tests and move services out of the network handlers constructors

marcamsn:


core/src/main/java/org/hisp/dhis/android/network/attribute/AttributeService.kt
```diff
@@ -26,22 +26,21 @@
...
-import org.hisp.dhis.android.core.arch.api.payload.internal.PayloadJackson
 import org.hisp.dhis.android.core.attribute.Attribute
+import org.hisp.dhis.android.network.common.HttpServiceClientKotlinx
 import org.koin.core.annotation.Singleton
 
 @Singleton
-internal class AttributeService(private val client: HttpServiceClient) {
-    suspend fun getAttributes(
+internal class AttributeService(private val client: HttpServiceClientKotlinx) {
+    suspend fun attributes(
```
taridepaco, line 39: Why removing the "get" from this method? In this case there is only a get request, but other times there will be post or delete

marcamsn: Yeah, I was just following what you did in `OptionService` to keep things consistent, but I did think while working on it that this might not be the clearest way. 

https://github.com/dhis2/dhis2-android-sdk/pull/2253/files#diff-d94cb9660272a03d5e53c7032f510f66a9468204fc04d092d831ef9099f18800

taridepaco (approved):

PR #2263, opened 2024-12-18
chore: [ANDROSDK-1959] Move CategoryCombo api logic to network 

marcamsn:
Solves [ANDROSDK-1959](https://dhis2.atlassian.net/browse/ANDROSDK-1959).

[ANDROSDK-1959]: https://dhis2.atlassian.net/browse/ANDROSDK-1959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/network/legendset/LegendSetDTOMapper.kt
```diff
@@ -36,15 +37,16 @@ internal fun legendSetDtoToDomainMapper(item: LegendSetDTO): LegendSet {
...
-    return Legend.builder()
-        .applyBaseIdentifiableFields(item)
-        .startValue(item.startValue)
-        .endValue(item.endValue)
-        .color(item.color)
-        .build()
+internal fun legendDtoToDomainMapper(item: LegendDTO, legendSetUid: String? = null): Legend {
+    return Legend.builder().apply {
+        applyBaseIdentifiableFields(item)
+        legendSetUid?.let { legendSet(ObjectWithUid.create(legendSetUid)) }
```
vgarciabnz, line 47: In the `main` code, this legendSetUid can't be empty. Actually, it comes from a `BaseIdentifiableObjectDTO` whose `uid` property is not nullable. The reason to make it nullable here is the test, isn't it? Couldn't we make it non-null in this method and just pass any string in the test case?

taridepaco (approved):

PR #2267, opened 2024-12-19
chore: [ANDROSDK-1956] migrate user model to network layer

taridepaco:
Related task: [ANDROSDK-1956](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-1956)

[ANDROSDK-1956]: https://dhis2.atlassian.net/browse/ANDROSDK-1956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/common/FeatureType.kt
```diff
@@ -25,39 +25,28 @@
...
  */
+package org.hisp.dhis.android.core.common
 
-package org.hisp.dhis.android.core.common;
-
-public enum FeatureType {
+import kotlinx.serialization.Serializable
 
+@Serializable
+enum class FeatureType(@JvmField val featureType: String, @JvmField val geometryType: String) {
```
vgarciabnz, line 33: These `@JvmField` annotations, are they required? It changes the Java API from `getFeatureType()` to just `featureType`.

core/src/main/java/org/hisp/dhis/android/core/common/FeatureType.kt
```diff
@@ -25,39 +25,28 @@
...
+
+    internal companion object {
+        @JvmStatic
+        fun valueOfFeatureType(featureType: String): FeatureType? {
+            for (value in entries) {
+                if (value.featureType == featureType || value.geometryType == featureType) {
+                    return value
+                }
             }
+            return null
```
vgarciabnz, line 49: This block could be simplify by using the `find()` function

core/src/main/java/org/hisp/dhis/android/core/common/FeatureType.kt
```diff
@@ -25,39 +25,28 @@
...
  *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+package org.hisp.dhis.android.core.common
 
-package org.hisp.dhis.android.core.common;
-
-public enum FeatureType {
+import kotlinx.serialization.Serializable
 
+@Serializable
```
vgarciabnz, line 32: There is a Serializable annotation in the `core` package, but I guess this is unavoidable if we don't want to duplicate the enums

vgarciabnz (changes requested):
The only comment I really want to discuss is the one about the `@JvmField` annotations; the rest are just comments.

core/src/main/java/org/hisp/dhis/android/core/common/FeatureType.kt
taridepaco: They are not necessary, I am fixing that so the api doesn't change.

taridepaco:
@vgarciabnz There are some scripts related with serializing Geometry that cannot be currently removed because this model is used for Events, which hasn't been converted yet.

vgarciabnz (approved):

PR #2273, opened 2025-01-09
chore: [ANDROSDK-1967] Move VisualizationService to network

vgarciabnz:
Solves [](https://dhis2.atlassian.net/browse/ANDROSDK-1967)

This service uses different methods depending on the DHIS2 version. The logic for this has been kept in the "core" package. We can decide if this makes sense or not. I think this is the first service in this situation.


core/src/main/java/org/hisp/dhis/android/core/visualization/internal/VisualizationCall.kt
```diff
@@ -50,44 +49,22 @@ internal class VisualizationCall(
...
-                        )
-                        PayloadJackson(listOf(visualization36.toVisualization()))
-                    } catch (ignored: Exception) {
-                        PayloadJackson()
-                    }
+            apiDownloader.downloadPartitioned(
+                uids,
+                MAX_UID_LIST_SIZE,
+                handler,
+            ) { partitionUids: Set<String> ->
```
taridepaco, line 57: I think this logic could be moved to the network layer. There, networkHandlerImpl would return a PayloadJson.

core/src/main/java/org/hisp/dhis/android/network/visualization/Visualization36DTO.kt
```diff
@@ -26,33 +26,46 @@
...
-    val code: String?,
-    val name: String?,
-    val displayName: String?,
-    val created: Date?,
-    val lastUpdated: Date?,
-    val deleted: Boolean?,
+@Serializable
+internal data class Visualization36DTO(
+    @SerialName("id") override val uid: String,
+    override val code: String? = BaseIdentifiableObjectDTO.CODE,
```
taridepaco, line 50: Since all these defaults are null. they can all be removed. In another PR I've removed the companion object of BaseIdentifiableObjectDTO

vgarciabnz: Sure, I will remove them

core/src/main/java/org/hisp/dhis/android/core/visualization/internal/VisualizationCall.kt
vgarciabnz: That was my question, I wasn't sure where to place this logic. I can move it to the network layer for this particular case, it might make more sense. In others, maybe not. It could be something to decide depending on the case

core/src/main/java/org/hisp/dhis/android/network/visualization/VisualizationDimensionDTO.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.visualization.LayoutPosition
+import org.hisp.dhis.android.core.visualization.VisualizationDimension
+import org.hisp.dhis.android.core.visualization.VisualizationDimensionItem
+
+@Serializable
+internal data class VisualizationDimensionDTO(
+    val id: String?,
+    val items: List<VisualizationDimensionItemDTO?>? = emptyList(),
```
taridepaco, line 39: can an element of the list be directly null?

taridepaco (changes requested):
Added some suggestions, and the ci failed due to the test VisualizationShould.kt

core/src/main/java/org/hisp/dhis/android/network/visualization/VisualizationDimensionDTO.kt
vgarciabnz: It seems it can. Check the file `visualizations_2.json`. I don't remember it, but it must be possible if we added it here.

core/src/main/java/org/hisp/dhis/android/core/arch/handlers/internal/ChildElementHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.arch.handlers.internal
+
+import org.hisp.dhis.android.core.common.CoreObject
+
+internal interface ChildElementHandler<O : CoreObject> {
```
vgarciabnz, line 32: This is the new handler I implemented for nested list that are a relation one-to-many. Before this change, we use the `LinkHandler<S, O>` handler which forces to pass a transform function. This transform function was typically used to inject the parent uid in the object, which now is done in the API mapper. So the transformation is not needed anymore. We could have kept the `LinkHandler` with the identity transformation, but I think it was clearer this way because the purpose is a bit different.

We can keep the `LinkHandler` for the classes that are actual links (those that transform to a `...Link` model).

@marcamsn @taridepaco

core/src/main/java/org/hisp/dhis/android/core/visualization/internal/VisualizationCall.kt
```diff
@@ -28,70 +28,29 @@
...
-                }
-            }
-        } else {
-            emptyList()
+        return apiDownloader.downloadPartitioned(
+            uids,
+            MAX_UID_LIST_SIZE,
+            handler,
+        ) { partitionUids: Set<String> ->
+            networkHandler.getVisualizations(partitionUids)
```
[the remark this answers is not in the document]
vgarciabnz: I can change it in the next PR (TrackerVisualization)

taridepaco (approved):

PR #2277, opened 2025-01-14
chore: [ANDROSDK-1973] Move TrackedEntityType api logic to network layer

taridepaco:
Related task:[ ANDROSDK-1973](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-1973)


Jenkinsfile
```diff
@@ -42,7 +50,11 @@ pipeline {
                 script {
                     echo 'Browserstack deployment and running tests'
                     sh 'chmod +x ./scripts/browserstackJenkins.sh'
-                    sh './scripts/browserstackJenkins.sh'
+                    retry(20) { // Retry up to 3 times
+                        timeout(time: 5, unit: 'MINUTES') {
```
vgarciabnz, line 54: I think 5 minutes is not enough for instrumented tests

vgarciabnz (changes requested):

Jenkinsfile
taridepaco: No, I changed retries instead of minutes, it should be 20

vgarciabnz (approved):

PR #2282, opened 2025-01-17
chore: [ANDROSDK-1983] Move DataValue to network layer

taridepaco:
Related task: ANDROSDK-1983[](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-1983)


core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueDTO.kt
```diff
@@ -0,0 +1,88 @@
...
+internal data class DataValueDTO(
+    override val deleted: Boolean?,
+    val dataElement: String,
+    val period: String,
+    @SerialName(DataValueFields.ORGANISATION_UNIT) val organisationUnit: String,
+    val categoryOptionCombo: String,
+    val attributeOptionCombo: String,
+    val value: String?,
+    val storedBy: String?,
+    val created: String,
```
vgarciabnz, line 47: `created` and `lastUpdated` are not nullable, is it right?

core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueImportSummaryDTO.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.imports.ImportStatus
+import org.hisp.dhis.android.core.imports.internal.DataValueImportSummary
+import org.hisp.dhis.android.network.common.dto.ImportConflictDTO
+import org.hisp.dhis.android.network.common.dto.ImportCountDTO
+
+@Serializable
+internal data class DataValueImportSummaryDTO(
+    val importCount: ImportCountDTO,
+    @SerialName("status") val importStatus: String,
```
vgarciabnz, line 41: Couldn't we just adapt the property name to avoid the definition of a different @SerialName ?

core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueImportSummaryWebResponseDTO.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.hisp.dhis.android.network.datavalue
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.imports.internal.DataValueImportSummaryWebResponse
+import org.hisp.dhis.android.network.common.dto.WebResponseDTO
+
+@Serializable
+internal data class DataValueImportSummaryWebResponseDTO(
```
vgarciabnz, line 36: This is an example of a class that was originally created for the API and whose properties map 1to1 to the domain class. I think it is required if we want to keep them separated, but I feel it like a kind of duplication

core/src/main/java/org/hisp/dhis/android/core/D2.kt
```diff
@@ -78,6 +79,10 @@ class D2 internal constructor(internal val d2DIComponent: D2DIComponent) {
         return d2DIComponent.httpServiceClient
     }
 
+    fun httpServiceClientKotlinx(): HttpServiceClientKotlinx {
```
vgarciabnz, line 82: Couldn't it be internal at least? It should avoid all the modifications in `HttpServiceClientKotlinx` done in this PR

core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueDTO.kt
taridepaco: They are non nullable in the domain model Builder, but nullable in the domain model definition. We need to check this.

core/src/main/java/org/hisp/dhis/android/network/common/dto/ImportConflictDTO.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.imports.internal.ImportConflict
+
+@Serializable
+internal data class ImportConflictDTO(
+    @SerialName("object") val objectName: String,
+    val value: String,
+    val errorCode: String? = null,
```
vgarciabnz, line 39: Could this defaults be omitted?

vgarciabnz (approved):
I leave it approved, my only comment was about some defaults (null) that could be omitted. There are a few more that I didn't commented

PR #2289, opened 2025-01-22
chore: [ANDROSDK-1965] Implement Koin Context Isolation in SDK

taridepaco:
Related task: [ANDROSDK-1965](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-1965)

[ANDROSDK-1965]: https://dhis2.atlassian.net/browse/ANDROSDK-1965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/d2/internal/D2DIComponentFactory.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/d2/internal/D2DIComponentFactory.kt
@@ -76,7 +76,6 @@ import org.hisp.dhis.android.core.visualization.VisualizationDIModule
 import org.hisp.dhis.android.core.wipe.internal.WipeDIModule
 import org.hisp.dhis.android.network.dependencyinjection.NetworkDIModule
 import org.koin.core.Koin
-import org.koin.core.component.KoinComponent
 import org.koin.dsl.koinApplication
 import org.koin.ksp.generated.module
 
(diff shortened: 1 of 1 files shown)


andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/arch/d2/internal/D2DIComponentFactory.kt
```diff
@@ -75,70 +75,86 @@ import org.hisp.dhis.android.core.validation.ValidationDIModule
...
-        }
+internal object D2DIComponentFactory : IsolatedKoinComponent {
 
-        return koinApp.koin.get<D2DIComponent>()
+    fun create(
+        d2Config: D2Configuration,
+        secureStore: SecureStore,
+        insecureStore: InsecureStore,
+    ): D2DIComponent {
+        DhisAndroidSdkKoinContext.koin = DhisAndroidSdkKoinContext.create(d2Config, secureStore, insecureStore).koin
```
vgarciabnz, line 157: I find this part a bit odd... It is assigning a property in an object and then using method declared in an interface (`getKoin()`), which internally refers to the same property in the that object. Couldn't it be simplified in some way? Or abstracted in some way?

PR #2292, opened 2025-01-23
chore: [ANDROSDK-1987] Move ValidationRuleService to api layer

taridepaco:
Related task: [ANDROSDK-1987](https://dhis2.atlassian.net/browse/ANDROSDK-1987)

[ANDROSDK-1987]: https://dhis2.atlassian.net/browse/ANDROSDK-1987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/validation/ValidationRule.java
```diff
@@ -52,57 +49,42 @@
...
 
-    @JsonProperty()
     public abstract Boolean skipFormValidation();
 
-    @JsonProperty()
     @ColumnAdapter(LeftValidationRuleExpressionColumnAdapter.class)
     public abstract ValidationRuleExpression leftSide();
 
-    @JsonProperty()
     abstract String leftSideExpression();
```
vgarciabnz, line 71: These auxiliary properties are suspicious... They are assigned a value in the builder as if both `leftSide()` and `leftSideExpression()` might or should have a value. It could happen that these are the legacy properties that were replaced by `leftSide()` (in this case, leftSide should be nullable in the DTO or have two differents DTO). Maybe we can look at the commit history to understand this and clean it up

core/src/main/java/org/hisp/dhis/android/core/validation/ValidationRule.java
```diff
@@ -68,22 +67,22 @@ public abstract class ValidationRule extends BaseNameableObject implements CoreO
...
-
-    abstract String leftSideDescription();
-
-    @ColumnAdapter(MissingValueStrategyColumnAdapter.class)
-    abstract MissingValueStrategy leftSideMissingValueStrategy();
+//    abstract String leftSideExpression();
+//
+//    abstract String leftSideDescription();
+//
+//    @ColumnAdapter(MissingValueStrategyColumnAdapter.class)
```
vgarciabnz, line 74: Remove these comments.
You could probably remove the class `MissingValueStrategyColumnAdapter` as well.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/network/validationrule/ValidationRuleExpressionDTO.kt
```diff
@@ -0,0 +1,48 @@
...
+
+package org.hisp.dhis.android.network.validationrule
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.validation.MissingValueStrategy
+import org.hisp.dhis.android.core.validation.ValidationRuleExpression
+
+@Serializable
+internal data class ValidationRuleExpressionDTO(
+    val expression: String,
```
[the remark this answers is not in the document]
taridepaco: How is this possible if the expression fields are not nullable in the domain model??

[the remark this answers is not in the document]
vgarciabnz: Maybe that is a potential failure in the existing model. I have double checked that it is possible to create a ValidationRule without specifying the expression or description.

We can either manage the null in the domain model or default to empty string in the DTO.

core/src/main/java/org/hisp/dhis/android/network/validationrule/ValidationRuleDTO.kt
```diff
@@ -62,26 +62,36 @@ internal data class ValidationRuleDTO(
...
     }
 }
 
 @Serializable
 internal class ValidationRulePayload(
     override val pager: PagerDTO?,
     @SerialName("validationRules") override val items: List<ValidationRuleDTO> = emptyList(),
-) : PayloadJson<ValidationRuleDTO>(pager, items)
+) : PayloadJson<ValidationRuleDTO>(pager, items) {
+    fun mapNotNullItems(
```
vgarciabnz, line 89: Would it make sense to add this method to `PayloadJson`?

taridepaco: I'd add it if we found another case were it is needed.

vgarciabnz (approved):

PR #2302, opened 2025-01-31
chore: [ANDROSDK-1997] Move OrganisationUnit to api layer

vgarciabnz:
Solves [ANDROSDK-1997](https://dhis2.atlassian.net/browse/ANDROSDK-1997).

The DTO for this model already existed, but the service was not moved. I also noticed a few things that were not right. They are commented in the code

[ANDROSDK-1997]: https://dhis2.atlassian.net/browse/ANDROSDK-1997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/androidTest/java/org/hisp/dhis/android/core/organisationunit/internal/OrganisationUnitCallMockIntegrationShould.kt
```diff
@@ -68,7 +69,7 @@ class OrganisationUnitCallMockIntegrationShould : BaseMockIntegrationTestEmptyEn
         val organisationUnits = listOf(orgUnit)
 
         // dependencies for the OrganisationUnitCall:
-        val organisationUnitService = OrganisationUnitService(d2.httpServiceClient())
+        val organisationUnitNetworkHandler: OrganisationUnitNetworkHandler = DhisAndroidSdkKoinContext.koin.get()
```
vgarciabnz, line 72: Useful way to get objects from the koin tree, so they don't have to have initialized in the tests. A lot of code could be replaced by this way of getting objects.

core/src/main/java/org/hisp/dhis/android/core/organisationunit/OrganisationUnit.java
```diff
@@ -182,21 +148,6 @@ public Builder closedDate(@NonNull String closedDateStr) throws ParseException {
...
         public abstract Builder displayNamePath(List<String> displayNamePath);
 
-        abstract OrganisationUnit autoBuild();
-
-        // Auxiliary fields to access values
-        abstract FeatureType featureType();
-        abstract String coordinates();
-        abstract Geometry geometry();
-
-        public OrganisationUnit build() {
```
vgarciabnz, line 192: This logic has been moved to the DTO.

core/src/main/java/org/hisp/dhis/android/network/common/dto/GeometryDTO.kt
```diff
@@ -29,18 +29,19 @@
...
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonElement
 import org.hisp.dhis.android.core.common.FeatureType
 import org.hisp.dhis.android.core.common.Geometry
 
 @Serializable
 internal data class GeometryDTO(
     val type: String?,
-    val coordinates: String?,
+    val coordinates: JsonElement?,
```
vgarciabnz, line 39: The format of this value is variable, it can be a list of double, a list of lists of doubles, or a list of lists of lists of double. There was a custom Serializer in the domain that took care of this. Here, it is modeled as a JsonElement and then it uses the `toString()` method. This class is used in tracker data objects, so will have to double test it there. Also, it will be required to have a `fromDomain()` method to serialize.

core/src/main/java/org/hisp/dhis/android/network/common/dto/GeometryDTO.kt
```diff
@@ -29,18 +29,19 @@
...
 ) {
     fun toDomain(): Geometry {
         return Geometry.builder()
-            .type(type?.let { FeatureType.valueOf(it) })
-            .coordinates(coordinates)
+            .type(type?.let { FeatureType.valueOfFeatureType(it) })
```
vgarciabnz, line 43: The format coming from the API is slightly different to the enum name, so it is necessary to use this auxiliary method.

core/src/main/java/org/hisp/dhis/android/network/organisationunit/OrganisationUnitDTO.kt
```diff
@@ -74,12 +77,37 @@ internal data class OrganisationUnitDTO(
...
             .programs(programs?.map { ObjectWithUid.create(it.uid) })
             .programs(programs?.map { ObjectWithUid.create(it.uid) })
             .dataSets(dataSets?.map { ObjectWithUid.create(it.uid) })
             .organisationUnitGroups(organisationUnitGroups?.map { it.toDomain() })
-            .displayNamePath(displayNamePath)
+            .displayNamePath(evaluateDisplayNamePath())
             .build()
     }
+
+    private fun evaluateDisplayNamePath(): List<String?>? {
```
vgarciabnz, line 89: The displayNamePath is evaluated here, so a lot classes can be removed from the domain.

core/src/main/java/org/hisp/dhis/android/core/organisationunit/OrganisationUnit.java
```diff
@@ -70,65 +61,40 @@ public enum Scope {
...
-    @JsonProperty()
     @ColumnAdapter(DbDateColumnAdapter.class)
     public abstract Date closedDate();
 
     @Nullable
-    @JsonProperty()
     public abstract Integer level();
 
     @Nullable
-    @JsonProperty()
```
vgarciabnz, line 96: coordinates, featureType and ancestors have been removed from the domain.

taridepaco (approved):

PR #2311, opened 2025-02-05
Support for login using two factor authentication

ifoche:
This code enables the SDK to provide support for 2FA enabled users.

It also fixes an existing but providing support for more than one cookie server side if present.


vgarciabnz:
This PR is closed and will be consulted for the development of the feature in the next release.

PR #2313, opened 2025-02-05
chore: [ANDROSDK-2002] Move SettingService to api layer

taridepaco:
Related task [ANDROSDK-2002](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2002)

[ANDROSDK-2002]: https://dhis2.atlassian.net/browse/ANDROSDK-2002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/network/settings/AnalyticsTeiWHONutritionDataDTO.kt
```diff
@@ -0,0 +1,52 @@
...
+
+package org.hisp.dhis.android.network.settings
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.settings.AnalyticsTeiWHONutritionData
+import org.hisp.dhis.android.core.settings.WHONutritionChartType
+
+@Serializable
+internal data class AnalyticsTeiWHONutritionDataDTO(
+    val teiSetting: String?,
```
vgarciabnz, line 37: This property does not come in the payload but it is inferred from the parent. It is currently managed in the handler. It happens the same with dataElement, indicators and attributes. We can keep it as it is for now.

core/src/main/java/org/hisp/dhis/android/network/settings/ProgramConfigurationSettingDTO.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package org.hisp.dhis.android.network.settings
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.settings.ProgramConfigurationSetting
+
+@Serializable
+internal data class ProgramConfigurationSettingDTO(
+    val uid: String?,
```
vgarciabnz, line 36: This property is not present in the payload. It is filled in `AppearanceSettingCall`. It can be removed here and just leave as it is; or it can passed from the parent in the method `toDomain()`

core/src/main/java/org/hisp/dhis/android/network/settings/DataSetConfigurationSettingDTO.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.hisp.dhis.android.network.settings
+
+import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.settings.DataSetConfigurationSetting
+
+@Serializable
+internal data class DataSetConfigurationSettingDTO(
+    val uid: String?,
```
vgarciabnz, line 36: The same as for `ProgramConfigurationSettingDTO`

core/src/main/java/org/hisp/dhis/android/network/settings/SynchronizationSettingsDTO.kt
```diff
@@ -0,0 +1,65 @@
...
+    val fileMaxLengthBytes: Int?,
+) {
+    fun toDomain(): SynchronizationSettings {
+        return SynchronizationSettings.builder()
+            .dataSync(dataSync?.let { DataSyncPeriod.from(it) })
+            .metadataSync(metadataSync?.let { MetadataSyncPeriod.from(it) })
+            .trackerImporterVersion(
+                trackerImporterVersion?.let { TrackerImporterVersion.valueOf(it) }
+                    ?: TrackerImporterVersion.V1,
+            )
```
vgarciabnz, line 55: This logic is not exactly equal to `@JsonEnumDefaultValue`. The annotation will assigned a default value when the enum is unknown: let's suppose the version is "V3", which is not present in the enum, the annotation will assign "V1" while the current logic wil fail (valueOf will throw an error).

I am not sure which one is more correct, but they are different

core/src/main/java/org/hisp/dhis/android/network/systemsettings/SystemSettingsDTO.kt
```diff
@@ -25,28 +25,47 @@
...
+@Serializable
+internal data class SystemSettingsDTO(
+    val keyFlag: String?,
+    val keyStyle: String?,
+    val keyDefaultBaseMap: String?,
+    val keyBingMapsApiKey: String?,
+) {
+    fun toDomain(): SystemSetting {
+        return SystemSetting.builder()
+            .key(keyFlag?.let { SystemSetting.SystemSettingKey.valueOf(it) })
```
vgarciabnz, line 43: What is the purpose of this method? the value of "keyFlag" can be something as "sierra_leone". It is trying to get the enum value from `SystemSetting.systemSettingKey` using the value

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/network/systemsettings/SystemSettingsDTO.kt
taridepaco: Since the two services are using the other toDomain functions, this default one is not used and can be removed

vgarciabnz (approved):

PR #2314, opened 2025-02-05
chore: [ANDROSDK-1983-2] Fix date format in DataValueDTO

vgarciabnz:
Fix a defect in ANDROSDK-1983. The serialization of dates must be done using the method `DateUtils.DATE_FORMAT.parse()` (the opposite to the method used in deserialization). The method `toString()` prints the date in a friendly format, such as:

image


vgarciabnz:
It is not important in this particular case because the properties `created` and `lastUpdated` are ignored by the server, but it could make it fail in other cases.

taridepaco (approved):

PR #2318, opened 2025-02-07
chore: [ANDROSDK-2014] Move Event to api layer

vgarciabnz:
Solves [ANDROSDK-2014](https://dhis2.atlassian.net/browse/ANDROSDK-2014).

Things to consider:
- Uses the approach `toDto()` as extension function instead of the function `fromDomain()`. It changes the existing `DataValueDTO.fromDomain()`.
- Adds extension functions to parse/format with DATE_FORMAT. It does not apply this extension function to existing code.
- Removes support for notes in version 2.29. There were still a few tests and properties even though the SDK does not support 2.29 anymore.

[ANDROSDK-2014]: https://dhis2.atlassian.net/browse/ANDROSDK-2014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/event/Event.java
marcamsn, line 1: Should we remove the JSON annotations from this file?

core/src/main/java/org/hisp/dhis/android/core/util/DateExtensions.kt
marcamsn, line 1: We already have some Date extensions on the DateUtils.kt file. Should we unify this?

```
    internal fun LocalDate.atStartOfDayInSystem(): Instant {
        return this.atStartOfDayIn(TimeZone.currentSystemDefault())
    }

    internal fun Date.toKtxInstant(): Instant {
        return Instant.fromEpochMilliseconds(this.time)
    }

    internal fun Instant.toJavaDate(): Date {
        return Date(this.toEpochMilliseconds())
    }
```

core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueDTO.kt
```diff
@@ -65,23 +65,21 @@ internal data class DataValueDTO(
...
+    return DataValueDTO(
+        deleted = this.deleted(),
+        dataElement = this.dataElement()!!,
+        period = this.period()!!,
+        orgUnit = this.organisationUnit()!!,
+        categoryOptionCombo = this.categoryOptionCombo()!!,
+        attributeOptionCombo = this.attributeOptionCombo()!!,
+        value = this.value(),
+        storedBy = this.storedBy(),
+        created = this.created()?.let { DateUtils.DATE_FORMAT.format(it) },
```
marcamsn, line 80: Should we use the date extensions created in this PR?

core/src/main/java/org/hisp/dhis/android/network/relationship/RelationshipDTO.kt
```diff
@@ -58,21 +58,19 @@ internal data class RelationshipDTO(
...
-                to = r.to()?.let { RelationshipItemDTO.fromDomain(it) },
-            )
-        }
-    }
+internal fun Relationship.toDto(): RelationshipDTO {
+    return RelationshipDTO(
+        deleted = this.deleted(),
+        relationship = this.uid(),
+        relationshipName = this.name(),
+        created = this.created()?.let { DateUtils.DATE_FORMAT.format(it) },
```
marcamsn, line 68: Same here

core/src/main/java/org/hisp/dhis/android/core/event/Event.java
vgarciabnz: We can't yet because they are used as nested properties in TrackedEntityInstances. They could be removed once TEI and enrollment are migrated

core/src/main/java/org/hisp/dhis/android/core/util/DateExtensions.kt
vgarciabnz: It makes sense. Or maybe moved them to a different file because they are all related to kotlinx.datetime

core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueDTO.kt
vgarciabnz: Yes, but I didn't apply the extension to existing files (event those modified in this PR). We can do it as a separate task to move them all.

core/src/main/java/org/hisp/dhis/android/network/relationship/RelationshipDTO.kt
vgarciabnz: Some thing, we can move them all in a separate task

core/src/main/java/org/hisp/dhis/android/network/common/dto/GeometryDTO.kt
```diff
@@ -45,3 +46,10 @@ internal data class GeometryDTO(
             .build()
     }
 }
+
+internal fun Geometry.toDto(): GeometryDTO {
+    return GeometryDTO(
+        type = this.type()?.geometryType,
+        coordinates = this.coordinates()?.let { parseToJsonElement(it) },
```
taridepaco, line 53: Maybe the jsonParser that we have declared here "import org.hisp.dhis.android.core.arch.json.internal.KotlinxJsonParser" can be used to serialize objects, so the config parameters are considered. At least, the "explicitNulls" one that skips null fields from the payload seems useful.

marcamsn (approved):

core/src/main/java/org/hisp/dhis/android/network/event/EventNetworkHandlerImpl.kt
```diff
@@ -0,0 +1,125 @@
...
+            program = eventQuery.commonParams.program,
+            startDate = eventQuery.getEventStartDate(),
+            paging = true,
+            page = eventQuery.page,
+            pageSize = eventQuery.pageSize,
+            lastUpdatedStartDate = eventQuery.lastUpdatedStr,
+            includeDeleted = true,
+            eventUid = eventQuery.getUidStr(),
+        )
+        return apiPayload.mapItems { it.toDomain() }
```
taridepaco, line 73: For consistency with the rest of the handlers, we can change later this to apiPayload.mapItems(EventPayload::toDomain)

core/src/main/java/org/hisp/dhis/android/network/relationship/RelationshipDTO.kt
taridepaco, line 68: We can use here the new toDateFormat, can be homogenized later in a dedicated task.

taridepaco (approved):

PR #2322, opened 2025-02-11
[ANDROSDK-2008] Move TrackedEntityInstance to api layer

vgarciabnz:
[ANDROSDK-2008](https://dhis2.atlassian.net/browse/ANDROSDK-2008)

- Moves TrackedEntityInstance to api layer.
- Removes Json annotations from nested classes (enrollement, note, etc), except for Geometry because it is used by the new importer as well. The new importer will be able to remove it once it is migrated.
- Adds a value class `ValueDTO` for the issue with the nulls in the API
- Modifies the common interface shared between the new and old importer, so @taridepaco, you can expect conflicts.
- There were some helper methods in this shared interface that have been moved to the TrackerAPIQuery object.
- SearchGrid objects are moved to network layer without correspondent object in domain. They are translated directly into TrackedEntityInstances

[ANDROSDK-2008]: https://dhis2.atlassian.net/browse/ANDROSDK-2008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz:
There is a test that have to be ignored until the migration of NewTracker is done

taridepaco (approved):

PR #2324, opened 2025-02-13
chore: [ANDROSDK-2018] Move DataSetCompleteRegistrations to api layer

marcamsn:
Solves [ANDROSDK-2018](https://dhis2.atlassian.net/browse/ANDROSDK-2018).

[ANDROSDK-2018]: https://dhis2.atlassian.net/browse/ANDROSDK-2018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationPostCall.kt
```diff
@@ -136,19 +131,13 @@ internal class DataSetCompleteRegistrationPostCall(
...
-                errorClass = DataValueImportSummaryWebResponse::class.java,
-            ) {
-                dataSetCompleteRegistrationService.postDataSetCompleteRegistrationsWebResponse(payload)
-            }.map { it.response }
-        } else {
-            coroutineAPICallExecutor.wrap {
-                dataSetCompleteRegistrationService.postDataSetCompleteRegistrations(payload)
-            }
+        return coroutineAPICallExecutor.wrap(
+            acceptedErrorCodes = listOf(HttpURLConnection.HTTP_CONFLICT),
```
marcamsn, line 137: Previously, `acceptedErrorCodes` only applied to versions `greaterOrEqual` than `V2_38`, but now it applies to all versions.  
Not sure if we want to keep it this way—it's not hard to revert, but it would make the code a bit more confusing,  
and we’d have to reimport the `versionManager` here.  

Another option would be to move `coroutineAPICallExecutors` to `NetworkHandler`, but I like that approach less. 😅  

Thoughts?

vgarciabnz: It is fine for to apply the errorCodes to both requests, although there might be cases where we want to explicitly define a different behavior based on the version. So I feel that we will have to use the coroutineAPICallExecutors in network in the future.

core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationPostCall.kt
```diff
@@ -136,19 +131,13 @@ internal class DataSetCompleteRegistrationPostCall(
...
+            errorClass = DataValueImportSummary::class.java,
```
vgarciabnz, line 138: This class changed from DataValueImportSummaryWebResponse to DataValueImportSummary. And it is not a DTO but a Jackson-annotated class. I think it will work because this class is not deserialized by Ktor but using the ObjectMapper in the error catcher. No need to change it now, but I think we have to think about this cases, maybe moving this part of the logic to the network layer

vgarciabnz: I have just seen the Jackson annotations for this class were removed, so it could failed to be deserialized.

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetCompleteRegistrationCall.kt
```diff
@@ -61,15 +61,14 @@ internal class DataSetCompleteRegistrationCall(
         )
 
         return partitions.flatMap { part ->
-            service.getDataSetCompleteRegistrations(
-                fields = DataSetCompleteRegistrationFields.allFields,
+            networkHandler.getDataSetCompleteRegistrations(
```
taridepaco, line 64: Maybe only `query.lastUpdatedStr` and  `part `could be passed and the rest could be defined in the handler implementation

marcamsn: I just updated it. ;)

vgarciabnz (approved):

PR #2333, opened 2025-02-19
chore: [ANDROSDK-2032] Refactor HttpClientService 

vgarciabnz:
Refactors done in this PR:
- HttpServiceClientKotlinx is renamed to HttpServiceClient and moved to `org.hisp.dhis.android.core.arch.api` (its original package). All the test using this service keeps working fine and there is nothing to change.
- The class PayloadJackson has been removed.
- The injected classes (ktor, okhttp, etc) are now applied to the new service.

Basically, the HttpServiceClientKotlinx takes the place of HttpServiceClient in all the senses.

This PR is on top of the one about DataStore.


.kotlin/sessions/kotlin-compiler-14817249305173960940.salive
marcamsn, line 1: I think we should git-ignore files in .kotlin/sessions

core/src/androidTest/java/org/hisp/dhis/android/realservertests/apischema/ApiSchemaService.kt
```diff
@@ -29,13 +29,12 @@
 package org.hisp.dhis.android.realservertests.apischema
 
 import org.hisp.dhis.android.core.arch.api.HttpServiceClient
-import org.hisp.dhis.android.core.arch.api.payload.internal.PayloadJackson
 import org.hisp.dhis.android.network.common.fields.Fields
 
 internal class ApiSchemaService(private val client: HttpServiceClient) {
```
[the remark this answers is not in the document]
vgarciabnz: It is in the "androidTest" folder, we have not created a "network" package for tests yet, we could do it

.kotlin/sessions/kotlin-compiler-14817249305173960940.salive
vgarciabnz: This PR updates the .gitignore as well. Here the file is just removed from git (it was added at some point)

taridepaco (approved):

PR #2335, opened 2025-02-21
chore: [ANDROSDK-2036] Replace BaseObjectShould  with BaseObjectKotlinxShould

taridepaco:
Related task [ANDROSDK-2036](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2036)

[ANDROSDK-2036]: https://dhis2.atlassian.net/browse/ANDROSDK-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/test/java/org/hisp/dhis/android/core/common/DataAccessShould.kt
```diff
@@ -25,28 +25,22 @@
...
-    public DataAccessShould() {
-        super("common/data_access.json");
-    }
-
-    @Override
+class DataAccessShould : BaseObjectKotlinxShould("common/data_access.json"), ObjectShould {
     @Test
-    public void map_from_json_string() throws IOException, ParseException {
-        DataAccess dataAccess = objectMapper.readValue(jsonStream, DataAccess.class);
+    @Throws(IOException::class, ParseException::class)
```
marcamsn, line 38: Should we keep the @Throws(IOException::class, ParseException::class) on kotlin if will not be used on Java?

core/src/test/java/org/hisp/dhis/android/core/common/AccessShould.kt
```diff
@@ -25,29 +25,23 @@
...
-import static com.google.common.truth.Truth.assertThat;
-
-public class AccessShould extends BaseObjectShould implements ObjectShould {
-
-    public AccessShould() {
-        super("common/access.json");
-    }
-
-    @Override
+class AccessShould : BaseObjectKotlinxShould("common/access.json"), ObjectShould {
```
marcamsn, line 36: If there are no more old BaseObjectShould uses, we could rename the BaseObjectKotlinxShould to BaseObjectShould. :)

taridepaco: There is still a couple of tests for exporting the db that use the old one and cannot be updated yet.

marcamsn (approved):

PR #2340, opened 2025-02-26
chore: [ANDROSDK-2020] Remove ObjectMapper uses from errorClass and errorCatcher

vgarciabnz:
Solves [ANDROSDK-2020](https://dhis2.atlassian.net/browse/ANDROSDK-2020).

It removes the parameter "errorClass" from the CoroutineApiCallExecutor and replaces it by "errorClassParser", which uses a DTO to deserialize the string and then it translates it to domain. The uses of the executor are moved to the network layer.

The separation between domain/network is not respected because the coroutineExecutor uses some network classes. Eventually the coroutineExecutor could be moved to the network layer.

There are a few tests pending to be added.


core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/HttpStatusCodes.kt
```diff
@@ -37,6 +37,7 @@ internal object HttpStatusCodes {
     const val REDIRECT_MAX = 399
 
     const val NOT_FOUND = 404
+    const val CONFLICT = 409
```
marcamsn, line 40: Why don't we better use the HttpStatusCode from io.ktor.http? :)

marcamsn (approved):

core/src/main/java/org/hisp/dhis/android/network/common/dto/HttpMessageResponseDTO.kt
```diff
@@ -43,4 +44,12 @@ internal data class HttpMessageResponseDTO(
             .applyWebResponseFields(this)
             .build()
     }
+
+    companion object {
```
taridepaco, line 48: Since all these toErrorClass functions are so similar, we could think about a way to abstract the common logic

core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryErrorCatcher.kt
```diff
@@ -31,9 +31,8 @@ import com.fasterxml.jackson.core.JsonProcessingException
 import com.fasterxml.jackson.databind.JsonMappingException
```
taridepaco, line 31: There are two jackson exceptions imported that may be changed to a kotlinx serialization equivalent

core/src/main/java/org/hisp/dhis/android/network/common/dto/HttpMessageResponseDTO.kt
vgarciabnz: I tried to create an interface or something like that, but I wasn't able to parametrized a type with the Serializable annotation (I need the `serializer()` method that is created by this annotation). I didn't manage to do it, but it might be another way

core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryErrorCatcher.kt
vgarciabnz: That's right, I will change it

PR #2346, opened 2025-03-04
chore: [ANDROSDK-2047] Refactor jackson object mapper with Kotlinx parser in db exporter classes

taridepaco:
Related task [ANDROSDK-2047](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2047)

[ANDROSDK-2047]: https://dhis2.atlassian.net/browse/ANDROSDK-2047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/common/BaseIdentifiableObject.java
```diff
@@ -106,11 +102,8 @@ public static String dateToDateStr(Date date) {
         return BaseIdentifiableObject.DATE_FORMAT.format(date);
     }
 
-    @JsonPOJOBuilder(withPrefix = "")
     public abstract static class Builder<T extends Builder> {
 
-        @JsonProperty(UID)
-        @JsonAlias({UUID})  // Introduced in 2.38 due to changes in userCredentials model DHIS2-12577
```
vgarciabnz, line 113: About this UUID, check this issue and the related conversation in slack (https://dhis2.atlassian.net/browse/ANDROSDK-1584). I am not fully sure if it can be removed or not, we probably need to check if it is fixed in a patch of 2.38

core/src/test/java/org/hisp/dhis/android/core/common/BaseObjectKotlinxShould.kt
```diff
@@ -39,6 +39,10 @@ open class BaseObjectKotlinxShould(private val jsonPath: String) {
         return deserializePath(jsonPath, serializer)
     }
 
+    protected fun <T> deserialize(serializer: kotlinx.serialization.KSerializer<T>, jsonString: String): T {
```
vgarciabnz, line 42: Not really important, but it would be good to keep consistency with the rest of the methods. For example the following method has the serializer as second argument and the name is more explicit. We can name them all `deserialize` with different parameters. Or make the more explicit, such as `deserializeString`. Up to you

core/src/main/java/org/hisp/dhis/android/core/arch/storage/internal/JsonKeyValueStoreImpl.kt
```diff
@@ -41,29 +47,56 @@ internal open class JsonKeyValueStoreImpl<O>(
 
     override fun set(o: O) {
         try {
-            val strObject = objectMapper().writeValueAsString(o)
+            val strObject = when (o) {
+                is DatabasesConfigurationOld ->
```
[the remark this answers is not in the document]
vgarciabnz: Maybe the constructor could accept a KSerializer<O> instead of a Class<O>

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/common/BaseIdentifiableObject.java
taridepaco: Wouldn't that alias be added as a JsonName in the dto?

core/src/test/java/org/hisp/dhis/android/core/common/BaseObjectKotlinxShould.kt
taridepaco: Ok, changing the order. I wrote it this way for consistency with all the decodeFromString and encodeToString methods of the parser, that require the serializer first. But since for the tests we have this wrapper, consistency within it is more important.

vgarciabnz: It would be ok to have the serializer first as well, we can modify both methods if it makes more sense

vgarciabnz (approved):

PR #2356, opened 2025-03-14
chore: [Androsdk-2056] networkplugins

taridepaco:
Related task [ANDROSDK-2056](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2056)

Added a new parameter to the D2Configuration for custom ktor plugins "networkplugins", analogously to the Okhttp ones.

[ANDROSDK-2056]: https://dhis2.atlassian.net/browse/ANDROSDK-2056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/D2Configuration.java
+++ b/core/src/main/java/org/hisp/dhis/android/core/D2Configuration.java
@@ -30,13 +30,17 @@ package org.hisp.dhis.android.core;
 
 import android.content.Context;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import com.google.auto.value.AutoValue;
+
+import org.hisp.dhis.android.core.common.Unit;
 
 import java.util.Collections;
 import java.util.List;
 
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
+import io.ktor.client.plugins.api.ClientPlugin;
 import okhttp3.Interceptor;
 
 @AutoValue
@@ -64,6 +68,9 @@ public abstract class D2Configuration {
     public abstract List<Interceptor> networkInterceptors();
 
     @NonNull
+    public abstract List<ClientPlugin<Unit>> networkplugins();
+
+    @NonNull
     public abstract Context context();
 
     public abstract Builder toBuilder();
@@ -74,7 +81,8 @@ public abstract class D2Configuration {
                 .connectTimeoutInSeconds(30)
                 .writeTimeoutInSeconds(30)
                 .networkInterceptors(Collections.emptyList())
-                .interceptors(Collections.emptyList());
+                .interceptors(Collections.emptyList())
+                .networkplugins(Collections.emptyList());
     }
 
     @AutoValue.Builder
@@ -95,6 +103,8 @@ public abstract class D2Configuration {
         public abstract Builder interceptors(List<Interceptor> interceptors);
 
         public abstract Builder networkInterceptors(List<Interceptor> networkInterceptors);
+
+        public abstract Builder networkplugins(List<ClientPlugin<Unit>> networkplugins);
 
         public abstract D2Configuration build();
     }
--- a/core/src/main/java/org/hisp/dhis/android/core/D2Configuration.java
+++ b/core/src/main/java/org/hisp/dhis/android/core/D2Configuration.java
@@ -35,12 +35,11 @@ import androidx.annotation.Nullable;
 
 import com.google.auto.value.AutoValue;
 
-import org.hisp.dhis.android.core.common.Unit;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/hisp/dhis/android/core/D2Configuration.java
```diff
@@ -63,6 +67,9 @@ public abstract class D2Configuration {
     @NonNull
     public abstract List<Interceptor> networkInterceptors();
 
+    @NonNull
+    public abstract List<ClientPlugin<Unit>> networkplugins();
```
vgarciabnz, line 71: We usually use cameCase for the method names (networkPlugins)

vgarciabnz, line 71: Wouldn't be better to set use `Any` instead of `Unit` for the type? We would have `ClientPlugin<Any>`.

Checking the ClientPlugin interface, it seems it is not restricted to any type. If we use `Unit`, we restrict the plugins to have type `Unit`, which I think could have some implications in the configuration they might have

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/KtorPlugins.kt
```diff
@@ -43,6 +42,7 @@ internal fun HttpClientConfig<*>.addKtorPlugins(
     install(ServerURLVersionRedirectionPlugin.instance)
     install(PreventURLDecodePlugin.instance)
     install(authenticator.instance)
-    install(HttpTimeout) {
+    for (plugin in d2Configuration.networkplugins()) {
+        install(plugin)
```
vgarciabnz, line 46: What happens if the plugin needs some configuration, such as `UserAgentPlugin`? Should the SDK accept a configuration block along with the plugin?

vgarciabnz (changes requested):

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/KtorPlugins.kt
```diff
@@ -42,7 +43,9 @@ internal fun HttpClientConfig<*>.addKtorPlugins(
     install(DynamicServerURLPlugin.instance)
     install(ServerURLVersionRedirectionPlugin.instance)
     install(PreventURLDecodePlugin.instance)
+    install(DefaultRequest) {}
```
vgarciabnz, line 46: Why the `HttpTimeout` plugin is replaced by the `DefaultRequest` plugin?

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/CustomPlugin.kt
```diff
@@ -0,0 +1,36 @@
...
+ *  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.api.internal
```
vgarciabnz, line 29: this package must not be internal

core/src/main/java/org/hisp/dhis/android/core/arch/api/internal/CustomPlugin.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import io.ktor.client.plugins.HttpClientPlugin
+
+class CustomPlugin<TBuilder : Any, TPlugin : Any>(
```
vgarciabnz, line 33: The plugin might not be custom but a built-in ktor plugin. What about `ClientPlugin`? or `AppPlugin`?

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2363, opened 2025-04-02
feat: [ANDROSDK-1747] Add support for /api/loginConfig endpoint

marcamsn:
This PR introduces full support for retrieving login configuration from the `/api/loginConfig` endpoint, including:

- `LoginConfig`, `LoginOidcProvider`, and `LoginPageLayout` data classes
- DTOs and network handler implementations for `loginConfig` response mapping
- Integration of the loginConfig call into `UserModule` (RxJava and blocking variants)
- Mock integration test for `LoginConfigCall`

This enables clients to dynamically retrieve login UI configuration, including OIDC providers, recovery password hide/show, layout options, and more.


vgarciabnz (approved):

core/src/main/java/org/hisp/dhis/android/core/user/loginconfig/LoginConfigObjectRepository.kt
```diff
@@ -0,0 +1,56 @@
...
+class LoginConfigObjectRepository internal constructor(
+    private val loginConfigCall: LoginConfigCall,
+    private val serverUrl: String,
+) : ReadOnlyObjectRepository<LoginConfig> {
+    override fun get(): Single<LoginConfig?> {
+        return rxSingle { loginConfigCall.download(serverUrl) }.map { it }
+    }
+
+    override fun blockingGet(): LoginConfig {
+        return runBlocking { loginConfigCall.download(serverUrl) }
```
vgarciabnz, line 44: I have tried running this method against a server lower than v41 and it throws this error. I don't think it is the error we would expect... Maybe we could throw a different error to avoid doing two requests (the first one to blockingExists and the second one to blockingGet). We could think about this when we define the login workflow
image

taridepaco (approved):

PR #2367, opened 2025-04-10
fix: [ANDROSDK-2067] Fix Sync State Handling for Deleted Unsynced Events + Migration for Existing Data

marcamsn:
This PR addresses a bug where deleting new events that were never synced make the TEI and Enrollment to remain in a "pending to sync" state, even though there was no actual data to sync. This results in the sync UI getting stuck in a loop, confusing users. [ANDROSDK-2067](https://dhis2.atlassian.net/browse/ANDROSDK-2067).

### Fixes included:
- Ensures `propagateState()` is always called after internal deletions, guaranteeing proper sync state updates for affected TEIs and Enrollments.
- Adds migration script (`171.sql`) to retroactively fix the aggregated sync state in existing databases for:
  - `Event`
  - `Enrollment`
  - `TrackedEntityInstance`

[ANDROSDK-2067]: https://dhis2.atlassian.net/browse/ANDROSDK-2067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz (approved):

vgarciabnz:
Good query, after a second reading it is clear to me. Good to know that this kind of queries can be done.

I noticed a small thing in the query, but I wouldn't change it because it can never happen: the queries contain something like this
```
WHERE e.aggregatedSyncState NOT IN (c.synced, c.relationship) 
AND e.syncState IN (c.synced, c.relationship)
```
the event sync state should be `AND e.syncState = c.synced` because the relationship events shouldn't be modified. But they are already filtered out in the previous filter `e.aggregatedSyncState NOT IN (c.synced, c.relationship)` because the relationship events have both the syncState and aggregatedSyncState to RELATIONSHIP, so there is no risk to accidentally pick one of them.

PR #2375, opened 2025-05-05
chore: [ANDROSDK-2074] create persistence layer scaffolding and first entities

taridepaco:
In this PR a new layer for the persistence responsibilities has been created, following a similar structure as in network. Inside there is a dependency injection package, common folder for some interfaces and common entities, and some other parallel packages as found in the domain layer.

Some already existing database access objects have been moved to the persistence layer, as soon they will be used in the mapping between domain and persistence models instead of by the ColumnAdapters

Related task [ANDROSDK-2074](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2074)

[ANDROSDK-2074]: https://dhis2.atlassian.net/browse/ANDROSDK-2074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/persistence/constant/ConstantDB.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.room.PrimaryKey
+import org.hisp.dhis.android.core.constant.Constant
+import org.hisp.dhis.android.core.util.dateFormat
+import org.hisp.dhis.android.persistence.common.BaseIdentifiableObjectDB
+import org.hisp.dhis.android.persistence.common.EntityDB
+import org.hisp.dhis.android.persistence.common.applyBaseIdentifiableFields
+
+@Entity
+internal data class ConstantDB(
+    @PrimaryKey(autoGenerate = true) val id: Long = 0,
```
vgarciabnz, line 41: Do you think we should have a default value here? I would say we should leave it nullable so it is autogenerated by the database when it is null.

Also, the column name should be "_id" instead of "id", either changing the property value or using the ColumnInfo.

Do you think we could add this property to EntityDB (in the same way it is in CoreObject right now)?

core/src/main/java/org/hisp/dhis/android/persistence/common/FilterQueryCriteriaDB.kt
```diff
@@ -0,0 +1,73 @@
...
+package org.hisp.dhis.android.persistence.common
+
+import org.hisp.dhis.android.core.arch.json.internal.KotlinxJsonParser
+import org.hisp.dhis.android.core.common.AssignedUserMode
+import org.hisp.dhis.android.core.common.FilterQueryCriteria
+import org.hisp.dhis.android.core.event.EventStatus
+import org.hisp.dhis.android.core.organisationunit.OrganisationUnitMode
+
+internal interface FilterQueryCriteriaDB {
+    val followUp: Boolean?
```
marcamsn, line 38: Should the followUp property be an `Int`?

core/src/main/java/org/hisp/dhis/android/persistence/configuration/DatabaseAccountImportDBDAO.kt
```diff
@@ -26,9 +26,11 @@
  *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package org.hisp.dhis.android.core.configuration.internal
+package org.hisp.dhis.android.persistence.configuration
 
 import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.configuration.internal.DatabaseAccountImportDB
```
marcamsn, line 32: Rename to `DatabaseAccountImport` so we can call this class `DatabaseAccountImportDB`.

core/src/main/java/org/hisp/dhis/android/persistence/constant/ConstantDB.kt
```diff
@@ -0,0 +1,73 @@
...
+    override val uid: String,
+    override val code: String?,
+    override val name: String?,
+    override val displayName: String?,
+    override val created: String?,
+    override val lastUpdated: String?,
+    override val deleted: Boolean?,
```
marcamsn, line 48: Boolean -> Int?

marcamsn (approved):

core/src/main/java/org/hisp/dhis/android/persistence/constant/ConstantDB.kt
taridepaco: We will remove the default 0 and replace this id with _id in an upcoming PR.

core/src/main/java/org/hisp/dhis/android/persistence/common/FilterOperatorsDB.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.hisp.dhis.android.core.arch.json.internal.KotlinxJsonParser
+import org.hisp.dhis.android.core.common.FilterOperators
+
+internal interface FilterOperatorsDB {
+    val le: String?
+    val ge: String?
+    val gt: String?
+    val lt: String?
+    val eq: String?
+    val inProperty: Set<String>?
```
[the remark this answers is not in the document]
taridepaco: Fixed in upcoming PR.

core/src/main/java/org/hisp/dhis/android/persistence/common/FilterQueryCriteriaDB.kt
taridepaco: It seems Room transforms Booleans to Int when writing to db and viceversa automatically, so it's OK to keep the Boolean here.

core/src/main/java/org/hisp/dhis/android/persistence/configuration/DatabaseAccountImportDBDAO.kt
taridepaco: This class is public, not sure if we can directly change it's name without making breaking changes.

core/src/main/java/org/hisp/dhis/android/persistence/constant/ConstantDB.kt
taridepaco: Same as other comment.

PR #2379, opened 2025-05-09
chore: [ANDROSDK-2078] Add `toDomain()` and `toDB()` mappers to Attribute, Category, DataElement and DataApproval persistance packages

marcamsn:
This PR closes [ANDROSDK-2078](https://dhis2.atlassian.net/browse/ANDROSDK-2078) by introducing missing mappers in the Attribute, Category, DataElement and DataApproval persistence packages.

  * Added `toDomain()` implementations in `Attribute`/`Category`/`DataElement`/`DataApproval` persistance packages to convert from DB entities to core models.
  * Added `toDB()` companion extensions on `Attribute`/`Category`/`DataElement`/`DataApproval` domain packages types to produce their corresponding `*DB` instances.

[ANDROSDK-2078]: https://dhis2.atlassian.net/browse/ANDROSDK-2078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/persistence/attribute/DataElementAttributeValueLinkDB.kt
```diff
@@ -38,4 +40,24 @@ internal data class DataElementAttributeValueLinkDB(
     val dataElement: String,
     val attribute: String,
     val value: String?,
-)
+) : EntityDB<DataElementAttributeValueLink> {
+
+    override fun toDomain(): DataElementAttributeValueLink {
+        return DataElementAttributeValueLink.builder()
+            .id(id?.toLong())
```
taridepaco, line 47: Is the id necessary on the domain?

taridepaco (approved):

core/src/main/java/org/hisp/dhis/android/persistence/dataelement/DataElementDB.kt
```diff
@@ -57,4 +65,50 @@ internal data class DataElementDB(
     val fieldMask: String?,
     val color: String?,
     val icon: String?,
```
vgarciabnz, line 67: What do you think about creating a `ObjectWithStyleDB` interface for these two fields? and adding an `applyStyleFields` method in the mapper? Or maybe it is too much duplication of classes in domain and db...

core/src/main/java/org/hisp/dhis/android/persistence/dataelement/DataElementDB.kt
```diff
@@ -57,4 +65,50 @@ internal data class DataElementDB(
...
+        return DataElement.builder().apply {
+            applyBaseNameableFields(this@DataElementDB)
+            id(id?.toLong())
+            valueType?.let { valueType(ValueType.valueOf(it)) }
+            zeroIsSignificant(zeroIsSignificant)
+            aggregationType(aggregationType)
+            formName(formName)
+            domainType(domainType)
+            displayFormName(displayFormName)
+            optionSet?.let { optionSet(ObjectWithUidDB(it).toDomain()) }
```
vgarciabnz, line 80: In Network we use sometimes the domain object directly, such as `ObjectWithUid.create(it)`. I don't have a preference, just wanted to mention because it wouldn't break the layer separation

vgarciabnz (approved):

core/src/main/java/org/hisp/dhis/android/persistence/attribute/DataElementAttributeValueLinkDB.kt
marcamsn: We will keep it for now, and remove it when we remove the field.

core/src/main/java/org/hisp/dhis/android/persistence/dataelement/DataElementDB.kt
marcamsn: Same, I thought on it, don't have a clear preference. I selected this one so it is consistent with other more complex objects. Would you suggest to avoid creating more objects when is not needed and use the domain object directly?

PR #2390, opened 2025-05-16
fix: [ANDROSDK-2085] Add TrackerVisualization sorting

vgarciabnz:
Solves [ANDROSDK-2085](https://dhis2.atlassian.net/browse/ANDROSDK-2085).

It adds and uses the `sorting` property in TrackerVisualization to sort the rows in the evaluation result.

It adds a method in TrackerLineListRepository to apply the sorting programatically if needed.

The property is added to the database as a serialized string.

[ANDROSDK-2085]: https://dhis2.atlassian.net/browse/ANDROSDK-2085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/analytics/trackerlinelist/internal/TrackerLineListSortingMapper.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private fun checkIsEnrollmentDate(
+        sortingItem: TrackerVisualizationSorting,
+        lineListItem: TrackerLineListItem.EnrollmentDate,
+    ): Boolean {
+        return sortingItem.dimension() == "enrollmentDate" ||
+            matchesRegexProgramUid(
+                sortingItem.dimension(),
+                "^(\\w{11})\\.enrollmentdate$".toRegex(),
```
taridepaco, line 79: Could this be a constant like UID_PATTERN or UID_REGEX_PATTERN?

taridepaco (approved):

PR #2394, opened 2025-05-22
chore: [ANDROSDK-2102] move appender logic of objectWithUidChildStore to stores

taridepaco:
The database logic for building the query and fetching the objects when appending children has been moved to their corresponding stores. This will allow transitioning to Room since Sqlite exclusive function were used.

Related task [ANDROSDK-2102](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2102)

[ANDROSDK-2102]: https://dhis2.atlassian.net/browse/ANDROSDK-2102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/legendset/internal/DataElementLegendSetLinkStoreImpl.kt
```diff
@@ -55,4 +57,12 @@ internal class DataElementLegendSetLinkStoreImpl(
             w.bind(3, o.sortOrder())
         }
     }
+
+    override fun getLinksForDataElement(dataElementUid: String): List<ObjectWithUid> {
+        val whereClause = WhereClauseBuilder()
+            .appendKeyStringValue(DataElementLegendSetLinkTableInfo.CHILD_PROJECTION.parentColumn, dataElementUid)
```
vgarciabnz, line 63: It would be good to get rid of these `CHILD_PROJECTION` properties if possible. The TableInfo classes will be generated by KSP, so it would be better not to generate these properties. As this function is in a very specific context, I would use the parent column (DataElement) directly and remove the CHILD_PROJECTION property.

We can discuss it in the daily and, if we finally change it, it could be changed in another PR.

vgarciabnz (approved):

PR #2395, opened 2025-05-23
chore: [ANDROSDK-2104] move appender logic of SingleParentChildStoreImpl to stores 

taridepaco:
The database logic for building the query and fetching the objects when appending children has been moved to their corresponding stores. This will allow transitioning to Room since Sqlite exclusive function were used.

Related task [ANDROSDK-2104](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2104)

[ANDROSDK-2104]: https://dhis2.atlassian.net/browse/ANDROSDK-2104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/programstageworkinglist/internal/ProgramStageWorkingListAttributeValueFilterStore.kt
```diff
@@ -32,4 +32,8 @@ import org.hisp.dhis.android.core.arch.db.stores.internal.ObjectWithoutUidStore
 import org.hisp.dhis.android.core.programstageworkinglist.ProgramStageWorkingListAttributeValueFilter
 
 internal interface ProgramStageWorkingListAttributeValueFilterStore :
-    ObjectWithoutUidStore<ProgramStageWorkingListAttributeValueFilter>
+    ObjectWithoutUidStore<ProgramStageWorkingListAttributeValueFilter> {
+    fun getProgramStageWorkingListAttributeValueFilterForProgramStageWorkingList(
```
vgarciabnz, line 36: As we are in the context of a `ProgramStageWorkingListAttributeValueFilter` store, maybe we could simplify the name of the function to `getForProgramStageWorkingList`.

And the same could apply to the comment above about the legends, maybe it is not needed to repeat the store element name.

vgarciabnz (approved):

PR #2415, opened 2025-06-09
[ANDROSDK-2114] allow null and not null filters for tracker line list

taridepaco:
In this PR, new filters have been added for elements in line list:
- Is null / empty -> filter": "EQ:NV",
- Is not null / empty -> filter": "NE:NV",

Related task[ ANDROSDK-2114](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2114)

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -843,6 +843,17 @@ public final class org/hisp/dhis/android/core/analytics/trackerlinelist/DataFilt
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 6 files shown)


vgarciabnz:
Haved you checked if this operator includes "empty" values in web? I mean, having the filter ("is" / "is not") null or empty. I guess SQLite will make a difference between a NULL and an empty string.

PR #2420, opened 2025-06-17
chore: [ANDROSDK-2120] Remove uses of _id

vgarciabnz:
This PR removes the uses of the property "_id" in CoreObject and the parameter "_id" in CoreColumns. The changes in "main" code are explained as comments.


core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntitySearchItemHelper.kt
```diff
@@ -50,7 +50,7 @@ object TrackedEntitySearchItemHelper {
             syncState = i.syncState(),
             aggregatedSyncState = i.aggregatedSyncState(),
             deleted = i.deleted() ?: false,
-            isOnline = i.id() == null,
+            isOnline = i.syncState() == null,
```
vgarciabnz, line 53: Checked that it is safe to use this property to tell apart the online instances.

core/src/main/java/org/hisp/dhis/android/core/relationship/internal/RelationshipDownloadAndPersistCallFactory.kt
```diff
@@ -164,7 +164,7 @@ internal class RelationshipDownloadAndPersistCallFactory(
             corruptedRelationships.addAll(relationshipStore.getRelationshipsByItem(builder.build()))
         }
         for (r in corruptedRelationships) {
-            relationshipStore.deleteById(r)
+            r.uid()?.let { relationshipStore.deleteIfExists(it) }
```
vgarciabnz, line 167: The `uid` is optional in the Relationship model, but it is actually optional. This is a legacy configuration from 2.29 where the Relationship objects didn't have a `uid`. Since 2.29 is not supported anymore, the relationship objects will always have a non-null `uid`.

We should clean up the Relationship model.

core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetInstanceSummarySQLStatementBuilder.kt
```diff
@@ -99,7 +98,6 @@ class DataSetInstanceSummarySQLStatementBuilder : DataSetInstanceSQLStatementBui
             "ELSE 5 END)"
 
         private val SELECT_CLAUSE = "SELECT " +
-            dot(DS_LIST_TABLE_ALIAS, DeletableDataColumns.ID) + AS + DeletableDataColumns.ID + "," +
```
vgarciabnz, line 102: Same as for DataSetInstance objects.

core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetInstanceSQLStatementBuilder.kt
```diff
@@ -207,7 +205,6 @@ open class DataSetInstanceSQLStatementBuilder : ReadOnlySQLStatementBuilder {
             joinDataSetCompleteRegistration
 
         private val INNER_SELECT_CLAUSE = "SELECT " +
-            DATAVALUE_ID + AS + DeletableDataColumns.ID + "," +
```
vgarciabnz, line 210: This is only used to assign a `_id` value to DataSetInstance objects. This model extends from CoreObject, but it doesn't map to a table. I have verified that the only consequence of this change is that the returned values have a null `_id` property.

core/src/main/java/org/hisp/dhis/android/core/arch/db/querybuilders/internal/SQLStatementBuilderImpl.kt
```diff
@@ -193,7 +192,7 @@ internal class SQLStatementBuilderImpl internal constructor(
         // TODO refactor to only generate for object without uids store.
         val whereClause =
             if (whereColumns.isEmpty()) {
-                CoreColumns.ID + " = -1"
```
vgarciabnz, line 196: The code should never enter this part of flow (check the comment above). This condition should always be false to avoid updating/deleting any property that has an empty "where" columns. The "0" has the same effect.

taridepaco (approved):

PR #2421, opened 2025-06-17
chore: [ANDROSDK-2121] Create store implementation and DAO Part 3

taridepaco:
In this PR, daos and store implementations for the new Room db apprach have been added to the entities in:
- datastore
- datavalue
- domain
- enrollment
- event
Related task: [ANDROSDK-2121](https://dhis2.atlassian.net/browse/ANDROSDK-2121)

[ANDROSDK-2121]: https://dhis2.atlassian.net/browse/ANDROSDK-2121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/persistence/common/querybuilders/ReadOnlySQLStatementBuilderImpl.kt
```diff
@@ -110,6 +111,16 @@ internal open class ReadOnlySQLStatementBuilderImpl(
         return RoomRawQuery("DELETE FROM $tableName WHERE $whereClause;")
     }
 
+    override fun updateWhere(updates: ArrayMap<String, Any>, whereClause: String): RoomRawQuery {
+        val setClause = updates.entries.joinToString(", ") { "${it.key} = ${formatValue(it.value)}" }
```
vgarciabnz, line 115: I think you would need to wrap the value with single quotes because it is a string literal. In the case of numbers, it wouldn't be required.

The class ContentValues handles this formatting automatically.

core/src/main/java/org/hisp/dhis/android/persistence/common/querybuilders/ReadOnlySQLStatementBuilder.kt
```diff
@@ -28,6 +28,7 @@
 
 package org.hisp.dhis.android.persistence.common.querybuilders
 
+import android.util.ArrayMap
```
vgarciabnz, line 31: Is it required to use the ArrayMap from Android? Couldn't we use just a Map?

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2431, opened 2025-06-24
chore: [ANDROSDK-2130] Remove databaseAdapter dependency in childrenAppenders

vgarciabnz:
Solves []().

It removes the databaseAdapter dependency in the ChildrenAppenders. Now, the ChlidrenAppenders fetch the stores from the koin tree.

Changes are mainly:
- Remove databaseAdapter as a property in all the repositories (object and collection).
- Modify the `create()` method in the ChildreAppenders to get the store from the koin tree.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -24,9 +24,9 @@
             <option value="$PROJECT_DIR$/buildSrc" />
             <option value="$PROJECT_DIR$/core" />
             <option value="$PROJECT_DIR$/instrumented-test-app" />
+            <option value="$PROJECT_DIR$/processor" />
           </set>
         </option>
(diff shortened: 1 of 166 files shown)


.idea/gradle.xml
```diff
@@ -24,9 +24,9 @@
             <option value="$PROJECT_DIR$/buildSrc" />
             <option value="$PROJECT_DIR$/core" />
             <option value="$PROJECT_DIR$/instrumented-test-app" />
+            <option value="$PROJECT_DIR$/processor" />
```
vgarciabnz, line 27: I am not sure about this change... This is not intentional I think

taridepaco (approved):

PR #2467, opened 2025-10-06
chore: [ANDROSDK-2153] Implement download by filters in the DownloadCall

taridepaco:
This PR enables downloading teis and events using used passed working lists, events and filters as parameters.

Other transaction related issues have been addressed; nested transactions have been simplified.

Related task [ANDROSDK-2153](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2153)

[ANDROSDK-2153]: https://dhis2.atlassian.net/browse/ANDROSDK-2153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
@@ -27,7 +27,7 @@
  */
 package org.hisp.dhis.android.core.arch.api.executors.internal
 
-import androidx.room.immediateTransaction
+import androidx.room.deferredTransaction
 import androidx.room.useWriterConnection
 import io.ktor.client.plugins.ClientRequestException
 import org.hisp.dhis.android.core.arch.api.internal.D2HttpException
@@ -127,7 +127,7 @@ internal class CoroutineAPICallExecutorImpl(
     ): P {
         return try {
             databaseAdapter.getCurrentDatabase().useWriterConnection { transactor ->
-                transactor.immediateTransaction {
+                transactor.deferredTransaction {
                     block().also { successfulTransactionRoom(cleanForeignKeyErrors) }
                 }
             }
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
@@ -27,7 +27,7 @@
  */
 package org.hisp.dhis.android.core.arch.api.executors.internal
 
-import androidx.room.deferredTransaction
+import androidx.room.immediateTransaction
 import androidx.room.useWriterConnection
 import io.ktor.client.plugins.ClientRequestException
 import org.hisp.dhis.android.core.arch.api.internal.D2HttpException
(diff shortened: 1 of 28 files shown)


core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackerQueryFactory.kt
```diff
@@ -51,7 +51,7 @@ internal abstract class TrackerQueryFactory<T, S : TrackerBaseSync>(
         val programSettings = programSettingsObjectRepository.blockingGet()
         val internalFactory = internalFactoryCreator.invoke(params, programSettings)
         lastUpdatedManager.prepare(programSettings, params)
-        return if (params.program() == null) {
+        return if (params.program() == null && params.programStageWorkingLists() == null) {
```
vgarciabnz, line 54: I think this condition have to check `trackedEntityInstanceFilters` and `eventFilters` as well. Maybe it can be abstracted in a method in `params`.

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackerQueryFactory.kt
```diff
@@ -67,8 +67,11 @@ internal abstract class TrackerQueryFactory<T, S : TrackerBaseSync>(
                         internalFactory.queryGlobal(globalPrograms)
                 }
             }
-        } else {
+        } else if (params.program() != null) {
             internalFactory.queryPerProgram(params.program())
+        } else {
+            val programs = params.programStageWorkingLists()?.map { it.program().uid() }
```
vgarciabnz, line 73: Same here, it should consider the program uids coming from `trackedEntityInstanceFilters`, `eventFilters` and `programStageWorkingLists`

core/src/main/java/org/hisp/dhis/android/core/arch/api/executors/internal/CoroutineAPICallExecutorImpl.kt
```diff
@@ -127,7 +127,7 @@ internal class CoroutineAPICallExecutorImpl(
     ): P {
         return try {
             databaseAdapter.getCurrentDatabase().useWriterConnection { transactor ->
-                transactor.immediateTransaction {
+                transactor.deferredTransaction {
```
vgarciabnz, line 130: Do we need `deferredTransaction` here? I was taking a quick look to `deferredTransaction` and I am not sure... maybe we can have a call for this.

core/src/main/java/org/hisp/dhis/android/core/event/internal/EventQueryBundleInternalFactory.kt
```diff
@@ -61,10 +63,21 @@ internal class EventQueryBundleInternalFactory(
...
         val builder = EventQueryBundle.builder()
+            .eventFilters(eventFilters.takeIf { it.isNotEmpty() } ?: programSettingFilters)
             .commonParams(commonParams)
 
-        return commonHelper.divideByOrgUnits(commonParams.orgUnitsBeforeDivision, commonParams.hasLimitByOrgUnit) {
+        return commonHelper.divideByOrgUnits(
+            commonParams.orgUnitsBeforeDivision,
+            commonParams.hasLimitByOrgUnit,
+        ) {
             builder.orgUnits(it).build()
```
vgarciabnz, line 81: In this logic we should support the case where the ASWA has filters defined but the user is not assigned (sharing settings) to any of them. I have the feeling that the list would be empty here and we can't tell if an empty list means no filters defined or filters defined but not assigned.

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2469, opened 2025-10-07
fix: [ANDROSDK-2132] Handle relationships deleted on server during sync

marcamsn:
* Detect “relationship not found/already deleted” cases (E4005/E4016/E4017) and delete the local relationship instead of leaving it in `TO_UPDATE`.
* Add `RelationshipNotFoundConflict` matcher (regex for both messages) with friendly display description.
* Refactor `RelationshipImportHandler` to handle a single relationship summary, check conflicts/description, and decide **Delete** vs **Update**; keep state propagation only when updating.
* Update `JobReportRelationshipHandler` to delete on the same error set.
* Tests: unit tests for conflict matcher + import handler; real integration test to verify a locally `TO_UPDATE` relationship is removed after server-side deletion.

[ANDROSDK-2132
](https://dhis2.atlassian.net/browse/ANDROSDK-2132)

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/relationship/internal/RelationshipImportHandler.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/relationship/internal/RelationshipImportHandler.kt
@@ -51,33 +51,49 @@ internal class RelationshipImportHandler internal constructor(
     ) {
         importSummaries?.filterNotNull()?.forEach { importSummary ->
             importSummary.reference()?.let { relationshipUid ->
-                val relationship = relationshipRepository.withItems().uid(relationshipUid).blockingGet()
-
-                val state = getSyncState(importSummary.status())
-
-                val relationshipNotFoundOnServer = checkRelationshipNotFoundOnServer(importSummary)
-
-                val handleAction = if (relationshipNotFoundOnServer) {
-                    relationship?.let { relationshipStore.deleteByEntity(it) }
-                    HandleAction.Delete
-                } else {
-                    val handledState =
-                        if (state == State.ERROR || state == State.WARNING) {
-                            State.TO_UPDATE
-                        } else {
-                            state
-                        }
-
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/hisp/dhis/android/core/relationship/internal/RelationshipImportHandler.kt
```diff
@@ -49,23 +51,61 @@ internal class RelationshipImportHandler internal constructor(
...
+        relationshipUid: String,
+    ): HandleAction {
+        val state = getSyncState(importSummary.status())
+        val handledState = if (state == State.ERROR || state == State.WARNING) {
+            State.TO_UPDATE
+        } else {
+            state
+        }
+
+        relationshipStore.setSyncStateOrDelete(relationshipUid, handledState)
```
vgarciabnz, line 95: I think this line can return a `HandleAction.Delete` if the relationship was successfully deleted. Shouldn't this whole method return that action?

core/src/main/java/org/hisp/dhis/android/core/relationship/internal/RelationshipImportHandler.kt
```diff
@@ -49,23 +51,61 @@ internal class RelationshipImportHandler internal constructor(
...
-                dataStatePropagator.propagateRelationshipUpdate(relationship)
-            }
+        val handleAction = if (relationshipNotFoundOnServer) {
+            handleRelationshipNotFound(relationship)
+        } else {
+            handleRelationshipUpdateState(importSummary, relationshipUid)
         }
 
-        processIgnoredRelationships(importSummaries, relationships)
+        if (handleAction != HandleAction.Delete) {
```
vgarciabnz, line 74: I think the propagator should be called always, even for deletions.

vgarciabnz (approved):

PR #2497, opened 2025-11-13
chore: [ANDROSDK-2183] Review online order clause in event query

taridepaco:
Column names for sorting in online queries were not being transforemed into the correct ones from the new tracker endpoint, as it was done for tei queries. Now that event queries are enabled, the mapping needed to be added as well.

Related task [ANDROSDK-2183](https://dhis2.atlassian.net/jira/software/c/projects/ANDROSDK/boards/154?selectedIssue=ANDROSDK-2183)

[ANDROSDK-2183]: https://dhis2.atlassian.net/browse/ANDROSDK-2183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryRepositoryScopeHelper.kt
```diff
@@ -137,7 +140,10 @@ internal class TrackedEntityInstanceQueryRepositoryScopeHelper(
             criteria.enrollmentStatus()?.let { builder.enrollmentStatus(listOf(it)) }
             criteria.enrolledAt()?.let { builder.programDate(it) }
             criteria.enrollmentOccurredAt()?.let { builder.incidentDate(it) }
-            criteria.order()?.let { applyOrder(builder, it) }
+            criteria.order()?.let {
+                val version = trackerParentCallHelper.getTrackerExporterVersion()
+                applyOrder(builder, it, version)
```
vgarciabnz: The TrackedEntityQuery call resolves the version in the network layer: it applies V1 or V2 depending on the endpoint. I think it is not required to resolve it at this point.
I would apply the same to Events. I think this is something that belongs to the network layer.

core/src/androidTest/java/org/hisp/dhis/android/core/MetadataCallRealIntegrationShould.kt
```diff
@@ -70,6 +71,225 @@ class MetadataCallRealIntegrationShould : BaseRealIntegrationTest() {
         // adding a new program..etc.
     }
 
+    //    @Test
+    fun test_event_filter_with_order_uses_correct_version() {
```
vgarciabnz: The MetadataCallRealIntegrationShould is used as a playground with an empty test method just to add code for quick tests. Shouldn't be better to add this long test case to its own test class (even if it is commented out)?

vgarciabnz (approved):

PR #2502, opened 2025-11-25
fix: [ANDROSDK-2194] Add hash-based unique database names

marcamsn:
This PR introduces hash-based database naming to guarantee uniqueness across accounts, even when server URLs normalize to the same base form.

**Unique Database Names**
- 	Adds an 8-char SHA-256 hash to database names: `processedUrl_username_hash_encrypted.db`
- 	Ensures no collisions between accounts pointing to similar URLs.
- 	Provides old format compatibility via getOldDatabaseName() for migration detection.

**New Migration301**
- 	Detecting accounts still using the old DB format.
- 	Renaming DB files.
- 	Migrating `sdk_resources` and `sdk_cache_resources` directories.
- 	Updating protectedDbName for pending imports.
- 	Updating stored configuration with new DB names.


core/src/main/java/org/hisp/dhis/android/core/configuration/internal/migration/Migration301.kt
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        val alreadyMigrated = configuration.accounts().any { account ->
+            account.databaseName().matches(Regex(".*_[0-9a-f]{8}_(encrypted|unencrypted)\\.db"))
+        }
+
+        if (alreadyMigrated) {
+            Log.d(TAG, "Database name hash migration already applied")
+            return
+        }
```
taridepaco: This guard clause returns if at least one db has the new name, but what about in there are other dbs with the old one? wouldn't the checking above be using all() instead of any()?

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/migration/Migration301.kt
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Updates the importDB configuration if the account has a pending import.
+     */
+    private fun updateImportDB(
+        account: DatabaseAccount,
+        oldDbName: String,
+        newDbName: String,
+    ): org.hisp.dhis.android.core.configuration.internal.DatabaseAccountImport? {
```
taridepaco: move the import to the top, when searching it's easy to have them all together.

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/migration/Migration301.kt
marcamsn: I think that if at least one migration has executed, the app has already applied all earlier migrations. Given that `any()` and `all()` as migration validations do the same, I opted for any to improve performance. :)

taridepaco (approved):

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/migration/Migration301.kt
```diff
@@ -0,0 +1,203 @@
...
+) {
+
+    fun apply() {
+        val configuration = databaseConfigurationStore.get()
+        if (configuration == null || configuration.accounts().isEmpty()) {
+            return
+        }
+
+        val alreadyMigrated = configuration.accounts().any { account ->
+            account.databaseName().matches(Regex(".*_[0-9a-f]{8}_(encrypted|unencrypted)\\.db"))
```
vgarciabnz: I think this regex will match if the username is 8 chars long with characters in that range, for example `servername_abcdabcd_encrypted.db`.

Couldn't we just assume that the migration has to run if this method has been called? There is a check in the caller function that verifies that the existing code is less that 301.

core/src/test/java/org/hisp/dhis/android/core/configuration/internal/DatabaseNameGeneratorShould.kt
```diff
@@ -38,42 +38,121 @@ class DatabaseNameGeneratorShould {
...
+        assertThat(name).matches("play-dhis2-org-android-current_user23_[0-9a-f]{8}_unencrypted\\.db")
+        assertThat(name).endsWith("_unencrypted.db")
+    }
+
+    @Test
+    fun return_same_name_for_http_and_https() {
+        val nameHttps = generator.getDatabaseName(url, username, true)
+        val nameHttp = generator.getDatabaseName(urlHttp, username, true)
+
+        assertThat(nameHttps).isEqualTo(nameHttp)
```
vgarciabnz, line 66: Maybe we can discuss this point, I am not sure if instances with different protocols should be considered the same.

PR #2503, opened 2025-11-25
chore: [ANDROSDK-2173] Handle FileResource status when posting files

taridepaco:
This PR implements batch verification for file resource storage status to prevent issues with files being wrongly assigned when posted without waiting for the STORED status.

**Key changes:**
- Added FileResourceStorageStatusVerifier with exponential backoff retry mechanism for batch verification
- Refactored file upload flows DataValueFileResourcePostCall, OldTrackerImporterFileResourcesPostCall and TrackerImporterFileResourcesPostCall to upload files first, verify in batch, then update values only for verified files
- Extracted helper functions to reduce complexity and improve code maintainability
- Added comprehensive test suite with 7 test cases covering various verification scenarios
- Introduced FileResourceUploadResult and FileResourceVerificationResult data classes to track upload and verification states

Related task: [ANDROSDK-2173](https://dhis2.atlassian.net/browse/ANDROSDK-2173)

[ANDROSDK-2173]: https://dhis2.atlassian.net/browse/ANDROSDK-2173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2004-2023, University of Oslo
+ *  Copyright (c) 2004-2025, University of Oslo
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
--- a/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
@@ -1,0 +1,182 @@
+/*
+ *  Copyright (c) 2004-2023, University of Oslo
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  Redistributions of source code must retain the above copyright notice, this
+ *  list of conditions and the following disclaimer.
+ *
+ *  Redistributions in binary form must reproduce the above copyright notice,
+ *  this list of conditions and the following disclaimer in the documentation
+ *  and/or other materials provided with the distribution.
+ *  Neither the name of the HISP project nor the names of its contributors may
+ *  be used to endorse or promote products derived from this software without
+ *  specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ *  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ *  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
```diff
@@ -0,0 +1,182 @@
...
+                    FileResourceStorageStatus.NONE -> {
+                        // Treat as pending
+                        Log.d(TAG, "File resource $uid has status NONE")
+                    }
+                }
+            }
+
+            // If there are still pending files and we haven't reached max attempts, wait before next poll
+            if (pendingUids.isNotEmpty() && attempt < maxAttempts) {
+                delay((delayMs * (2.0.pow(attempt) - 1)).toLong())
```
marcamsn: I think there's a critical bug in the exponential backoff calculation:

`delay((delayMs * (2.0.pow(attempt) - 1)).toLong())`

This formula grows exponentially without bounds:
- Attempt 10: ~17 minutes
- Attempt 20: ~12 days
- Attempt 30: ~34 years

With DEFAULT_MAX_ATTEMPTS = 30, this will cause the app to freeze indefinitely when files fail verification.

Maybe we could do a Linear backoff `delay(delayMs * attempt)`or cap the exponential backoff with a maxDelay like this:

```
val maxDelay = 10_000L
delay(minOf(delayMs * (2.0.pow(attempt - 1)).toLong(), maxDelay))
```

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
```diff
@@ -0,0 +1,182 @@
...
+        )
+
+        return results
+    }
+
+    /**
+     * Verifies a single batch of file resources.
+     */
+    private suspend fun verifyBatch(uids: List<String>): Map<String, FileResourceVerificationResult> {
+        return uids.associateWith { uid ->
```
marcamsn: I think the `verifyBatch` processes files sequentially rather than in parallel. With each verification taking 1-2 seconds, a batch of 10 files takes 10-20 seconds instead of a couple of seconds.

Should we try something like this?

```
private suspend fun verifyBatch(uids: List<String>): Map<String, FileResourceVerificationResult> {
    return coroutineScope {
        uids.map { uid ->
            async { uid to verifyFileResource(uid) }
        }.awaitAll().toMap()
    }
}
```

taridepaco: I fact, it is possible to request multiple file resources on a single query. Let me change it

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceStorageStatusVerifier.kt
```diff
@@ -0,0 +1,197 @@
...
+
+        for (attempt in 1..maxAttempts) {
+            if (pendingUids.isEmpty()) {
+                break
+            }
+
+            Log.d(TAG, "Verification attempt $attempt/$maxAttempts for ${pendingUids.size} pending files")
+
+            // Process in batches to avoid overwhelming the server
+            val currentBatch = pendingUids.take(batchSize)
```
vgarciabnz, line 84: This method verifies the uids in batches (default size 10) with a maximum of 5 attempts by default. It seems that if the uid list size is, for example, 60 uids, this method will only process the first 50 uids.

taridepaco:
Due to a bug on the server it is not possible to verify if the file resources have been successfully stored, so this PR would worsen the current file upload behavior.

PR #2513, opened 2025-12-11
fix: [ANDROSDK-2214] Handle break-the-glass for DHIS2 v42 servers returning 404

marcamsn:
In DHIS2 v42+, the endpoint `/tracker/trackedEntities/{uid}?program={programUid}` returns 404 instead of 403 to hide ownership information from unauthorized users.

Now the SDK detects break-the-glass scenarios when a 404 is returned:
1. When a 404 is received with a program filter, makes a second request without the program.
2. Checks whether the TEI includes a programOwner for the requested program.
3. Verifies whether the owner org unit is within `SEARCH` scope but not within `CAPTURE` scope.
4. If these conditions are met, return an `OWNERSHIP_ACCESS_DENIED` error to trigger the break-the-glass flow.

This is fully backward-compatible with older servers that still return 403.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
@@ -27,11 +27,13 @@
  */
 package org.hisp.dhis.android.core.trackedentity.internal
 
+import io.ktor.http.HttpStatusCode
 import org.hisp.dhis.android.core.arch.api.executors.internal.CoroutineAPICallExecutor
 import org.hisp.dhis.android.core.arch.api.payload.internal.Payload
 import org.hisp.dhis.android.core.arch.handlers.internal.IdentifiableDataHandlerParams
 import org.hisp.dhis.android.core.arch.helpers.Result
 import org.hisp.dhis.android.core.maintenance.D2Error
+import org.hisp.dhis.android.core.maintenance.D2ErrorCode
 import org.hisp.dhis.android.core.program.internal.ProgramDataDownloadParams
 import org.hisp.dhis.android.core.relationship.internal.RelationshipDownloadAndPersistCallFactory
 import org.hisp.dhis.android.core.relationship.internal.RelationshipItemRelatives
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
```diff
@@ -133,19 +138,68 @@ internal class TrackedEntityInstanceDownloadCall(
...
+            }
+            else -> result
+        }
+    }
+
+    private suspend fun checkOwnershipOnNotFound(
+        uid: String,
+        query: TrackerAPIQuery,
+    ): Result<TrackedEntityInstance?, D2Error> {
+        val program = query.commonParams.program ?: return Result.Success(null)
```
vgarciabnz: I think that it should return the original error if the program is not defined (instead of null).

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
```diff
@@ -133,19 +138,68 @@ internal class TrackedEntityInstanceDownloadCall(
...
+        val queryWithoutProgram = TrackerAPIQuery(
+            commonParams = query.commonParams.copy(program = null),
+        )
+
+        val teiResult = coroutineCallExecutor.wrap(storeError = false) {
+            trackerCallFactory.getTrackedEntityCall().getEntityCall(uid, queryWithoutProgram)
+        }
+
+        return when (teiResult) {
+            is Result.Failure -> Result.Success(null)
```
vgarciabnz: why this catch? shouldn't it return the error?

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
```diff
@@ -133,19 +138,68 @@ internal class TrackedEntityInstanceDownloadCall(
...
+                ) {
+                    Result.Failure(
+                        D2Error.builder()
+                            .errorCode(D2ErrorCode.OWNERSHIP_ACCESS_DENIED)
+                            .errorDescription("OWNERSHIP_ACCESS_DENIED")
+                            .httpErrorCode(HttpStatusCode.NotFound.value)
+                            .build(),
+                    )
+                } else {
+                    Result.Success(null)
```
vgarciabnz: same here, I think it should return the original error?

core/src/main/java/org/hisp/dhis/android/core/trackedentity/internal/TrackedEntityInstanceDownloadCall.kt
```diff
@@ -133,19 +138,68 @@ internal class TrackedEntityInstanceDownloadCall(
...
         }
+
+        val result = coroutineCallExecutor.wrap(
+            storeError = true,
+            errorCatcher = TrackedEntityInstanceCallErrorCatcher(),
+        ) {
+            trackerCallFactory.getTrackedEntityCall().getEntityCall(uid, query)
+        }
+
+        return when {
```
vgarciabnz: This block, isn't equivalent to check for Failure & NotFound code, and everything else return `result`? There is only one case that is managed differently.

marcamsn: Yes, I'll fix it

vgarciabnz (approved):

PR #2514, opened 2025-12-12
fix: [ANDROSDK-2213] Enrollment trackedEntity attributes are incorrectly marked as SYNCED

taridepaco:
This PR fixes a bug where all tei attributes (both from the tei type and the program) were set as SYNCED when the tei was checked to be posted in the server, which is strictly incorrect since the program related attributes are sent in the enrollment part of the post. This has been solved by splitting the SYNCED setting attributes by enrollment and tei job result fetching.

Related task: [ANDROSDK-2213](https://dhis2.atlassian.net/browse/ANDROSDK-2213)

[ANDROSDK-2213]: https://dhis2.atlassian.net/browse/ANDROSDK-2213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/tracker/importer/internal/JobReportTrackedEntityHandler.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/tracker/importer/internal/JobReportTrackedEntityHandler.kt
@@ -87,9 +87,8 @@ internal class JobReportTrackedEntityHandler internal constructor(
 
             if (tetAttributes.isNotEmpty()) {
                 trackedEntityAttributeValueStore.setSyncStateByAttributes(uid, tetAttributes, State.SYNCED)
+                trackedEntityAttributeValueStore.removeDeletedAttributeValuesByInstanceAndAttributes(uid, tetAttributes)
             }
         }
-
-        trackedEntityAttributeValueStore.removeDeletedAttributeValuesByInstance(uid)
     }
 }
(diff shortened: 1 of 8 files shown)


marcamsn (approved):

core/src/main/java/org/hisp/dhis/android/core/tracker/importer/internal/JobReportTrackedEntityHandler.kt
```diff
@@ -76,7 +78,18 @@ internal class JobReportTrackedEntityHandler internal constructor(
...
+        if (teType != null) {
+            val tetAttributes = trackedEntityTypeAttributeStore.getForTrackedEntityType(teType)
+                .mapNotNull { it.trackedEntityAttribute()?.uid() }
+
+            if (tetAttributes.isNotEmpty()) {
+                trackedEntityAttributeValueStore.setSyncStateByAttributes(uid, tetAttributes, State.SYNCED)
+            }
+        }
+
         trackedEntityAttributeValueStore.removeDeletedAttributeValuesByInstance(uid)
```
vgarciabnz: Shouldn't we apply the same logic for Enrollments? A successful synchronization might be caused by the deletion of an attribute, and the attribute could be linked to the enrollment

vgarciabnz (approved):

PR #2524, opened 2025-12-29
chore: [ANDROSDK-2161] add last access timestamp to accounts

taridepaco:
This PR adds a new last access timestamp to user accounts. This is stored on the device so no login is required.
The parameter is updated when logging in, logging out and also for the current account when requesting all or just the current one.
Related task: [ANDROSDK-2161](https://dhis2.atlassian.net/browse/ANDROSDK-2161)

[ANDROSDK-2161]: https://dhis2.atlassian.net/browse/ANDROSDK-2161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/androidTest/java/org/hisp/dhis/android/testapp/user/AccountManagerMockIntegrationShould.kt
+++ b/core/src/androidTest/java/org/hisp/dhis/android/testapp/user/AccountManagerMockIntegrationShould.kt
@@ -237,6 +237,55 @@ class AccountManagerMockIntegrationShould : BaseMockIntegrationTestEmptyEnqueabl
         loginAndDeleteAccount(user2, pass2, server2)
     }
 
+    @Test
+    fun update_last_access_date_when_getting_accounts() {
+        if (d2.userModule().blockingIsLogged()) {
+            d2.userModule().blockingLogOut()
+        }
+
+        dhis2MockServer.enqueueLoginResponses()
+        d2.userModule().blockingLogIn(user1, pass1, dhis2MockServer.baseEndpoint)
+
+        val initialAccount = d2.userModule().accountManager().getCurrentAccount()
+        val initialLastAccessDate = initialAccount?.lastAccessDate()
+        assertThat(initialLastAccessDate).isNotNull()
+
+        Thread.sleep(1000)
+
+        val accounts = d2.userModule().accountManager().getAccounts()
+        val updatedAccount = accounts.find { it.username() == user1 }
+        val updatedLastAccessDate = updatedAccount?.lastAccessDate()
+
+        assertThat(updatedLastAccessDate).isNotNull()
+        assertThat(updatedLastAccessDate).isNotEqualTo(initialLastAccessDate)
+
+        loginAndDeleteAccount(user1, pass1, dhis2MockServer)
+    }
+
+    @Test
+    fun update_last_access_date_when_getting_current_account() {
+        if (d2.userModule().blockingIsLogged()) {
+            d2.userModule().blockingLogOut()
+        }
+
+        dhis2MockServer.enqueueLoginResponses()
+        d2.userModule().blockingLogIn(user1, pass1, dhis2MockServer.baseEndpoint)
+
+        val initialAccount = d2.userModule().accountManager().getCurrentAccount()
+        val initialLastAccessDate = initialAccount?.lastAccessDate()
+        assertThat(initialLastAccessDate).isNotNull()
+
+        Thread.sleep(1000)
+
+        val updatedAccount = d2.userModule().accountManager().getCurrentAccount()
+        val updatedLastAccessDate = updatedAccount?.lastAccessDate()
+
+        assertThat(updatedLastAccessDate).isNotNull()
+        assertThat(updatedLastAccessDate).isNotEqualTo(initialLastAccessDate)
+
+        loginAndDeleteAccount(user1, pass1, dhis2MockServer)
+    }
+
     private fun addDataValue() {
         val period = d2.periodModule().periodHelper().blockingGetPeriodForPeriodTypeAndDate(PeriodType.Yearly, Date())
         val orgunit = d2.organisationUnitModule().organisationUnits().one().blockingGet()!!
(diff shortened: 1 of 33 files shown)


core/src/main/java/org/hisp/dhis/android/core/configuration/internal/DatabaseAccount.java
```diff
@@ -50,6 +50,9 @@ public abstract class DatabaseAccount {
     @NonNull
     public abstract String databaseCreationDate();
 
+    @Nullable
+    public abstract String lastAccessDate();
```
marcamsn: Should we return an Instant instead of a String?

marcamsn: I guess not since we have the same in the databaseCreationDate and we would need to have a custom serializer for this.

marcamsn (commented):
The `updateLastAccessDate` logic is more or less duplicated here:
- `AccountManagerImpl.updateCurrentAccountLastAccessDate()`
- `LogOutCall.updateLastAccessDate()`
- `RoomMultiUserDatabaseManager.updateLastAccessDate()`

Maybe we could consider extracting this logic into a single method in a class or in `DatabaseConfigurationHelper` so we centralize it in one place. Wdyt?

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/DatabaseAccount.java
taridepaco: Yes, I used a similar approach as in databaseCreationDate

taridepaco:
Ok, the logic of the wrappers can be simplified, the get and set of the configuration store can be moved to the main function.

marcamsn (approved):

vgarciabnz:
There is not any change in `DatabaseAccountDB`, is it actually persisted in the shared preferences? I think the test classes only check the value for the current account, which is evaluated on-the-fly.

vgarciabnz:
I thought we wanted to change the type of `databaseCreationDate` to a Date as well, just to keep them both with the same value type. I will be a breaking change for the App though. It is only used in a test class in the App, just the builder.

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/DatabaseConfigurationHelper.kt
```diff
@@ -70,7 +73,8 @@ internal class DatabaseConfigurationHelper(
             .serverUrl(serverUrl)
             .databaseName(dbName)
             .encrypted(encrypt)
-            .databaseCreationDate(dateProvider.dateStr)
+            .databaseCreationDate(DateUtils.DATE_FORMAT.parse(dateProvider.dateStr))
+            .lastAccessDate(DateUtils.DATE_FORMAT.parse(dateProvider.dateStr))
```
vgarciabnz, line 76: Maybe we can review the use of `DateProvider` here. The provider is basically returning a string value as a result of parsing a `Date()` object, and here it is converted back from string to Date.

I think the purpose of the `DateProvider` was to be able to mock the current date. We have the `ClockProvider` class that has the same purpose, we would probably use it here. It can be a separate task if needed.

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/DatabaseConfigurationHelper.kt
```diff
@@ -91,6 +95,33 @@ internal class DatabaseConfigurationHelper(
...
+
+    private fun updateAccountLastAccessDate(
+        configuration: DatabasesConfiguration?,
+        serverUrl: String,
+        username: String,
+    ): DatabasesConfiguration? {
+        val account = getAccount(configuration, serverUrl, username) ?: return configuration
+
+        val updatedAccount = account.toBuilder()
+            .lastAccessDate(Date())
```
vgarciabnz: Here, the dateProvider is not used. I don't know if it should be used, but it should be consistent with the other usage.

core/src/main/java/org/hisp/dhis/android/persistence/configuration/DatabaseAccountDB.kt
```diff
@@ -66,7 +70,8 @@ internal fun DatabaseAccount.toDB(): DatabaseAccountDB {
         username = username(),
         serverUrl = serverUrl(),
         databaseName = databaseName(),
-        databaseCreationDate = databaseCreationDate(),
+        databaseCreationDate = databaseCreationDate().dateFormat()!!,
```
vgarciabnz: Would it make sense to create a nonNull version of this extension? If the caller object (Date) is not null, the return value type may be nonNull

core/src/androidTest/java/org/hisp/dhis/android/testapp/user/AccountManagerMockIntegrationShould.kt
```diff
@@ -237,6 +237,55 @@ class AccountManagerMockIntegrationShould : BaseMockIntegrationTestEmptyEnqueabl
         loginAndDeleteAccount(user2, pass2, server2)
     }
 
+    @Test
```
vgarciabnz, line 240: Could we add a test with two accounts? And then verify that both of them have a nonNull lastAccessDate

core/src/main/java/org/hisp/dhis/android/core/configuration/internal/DatabaseConfigurationHelper.kt
taridepaco: I agree, the DateProvider is only used for this purpose and may be rethought. Probably doesn't even need to be a Singleton.

core/src/main/java/org/hisp/dhis/android/persistence/configuration/DatabaseAccountDB.kt
taridepaco: Creating nonNull versions of these methods and checking all usage.

vgarciabnz (approved):

PR #2527, opened 2026-01-13
feature: [ANDROSDK-2193] Add dataSet support to DataValue for DHIS2 v43+ compatibility

marcamsn:
Adds `sourceDataSet` field to DataValue to support the mandatory dataSet parameter required by DHIS2 v43+.

**Changes:**
- Add `sourceDataSet` field to DataValue model
- Add new `value()` method with dataSet parameter
- Deprecate old `value()` method with `@LegacyDataValueApi` annotation
- Download and upload DataValues per DataSet
- Add migration to populate sourceDataSet for existing data
- Marks v2.43 as supported


core/src/main/java/org/hisp/dhis/android/network/datavalue/DataValueNetworkHandlerImpl.kt
```diff
@@ -49,22 +49,28 @@ internal class DataValueNetworkHandlerImpl(
...
+            dataSetUids = dataSetUid,
             periodIds = commaSeparatedCollectionValues(bundle.periodIds),
             orgUnitUids = commaSeparatedCollectionValues(bundle.rootOrganisationUnitUids),
             children = true,
             paging = false,
             includeDeleted = true,
         )
-        return apiResponse.dataValues.map { it.toDomain() }
+        return apiResponse.dataValues.map { dto ->
+            DataValueInternalAccessor.insertSourceDataSet(dto.toDomain().toBuilder(), dataSetUid).build()
```
taridepaco: Isn't this done already in the toDomain() of the dto?

marcamsn: I'll fix this. :)

vgarciabnz (approved):

PR #2533, opened 2026-01-21
fix: [ANDROSDK-2236] Use DataInputPeriods to determine editableStatus in DataSetInstanceService

taridepaco:
This PR adds an editable status check for the data input period. Also adds a new DataSetNonEditableReason enum item for this, to which the app will need to adapt.

Related task [ANDROSDK-2236](https://dhis2.atlassian.net/browse/ANDROSDK-2236)

[ANDROSDK-2236]: https://dhis2.atlassian.net/browse/ANDROSDK-2236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetInstanceServiceImpl.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetInstanceServiceImpl.kt
@@ -321,8 +321,7 @@ internal class DataSetInstanceServiceImpl(
                     val openingDate = dataInputPeriod.openingDate()
                     val closingDate = dataInputPeriod.closingDate()
 
-                    openingDate == null && closingDate == null ||
-                        (openingDate?.let { !currentDate.before(it) } ?: true) &&
+                    (openingDate?.let { !currentDate.before(it) } ?: true) &&
                         (closingDate?.let { !currentDate.after(it) } ?: true)
                 }
             } ?: false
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/hisp/dhis/android/core/dataset/internal/DataSetInstanceServiceImpl.kt
```diff
@@ -306,6 +309,25 @@ internal class DataSetInstanceServiceImpl(
...
+        return dataInputPeriods.isNullOrEmpty() || dataInputPeriods
+            .filter { it.period().uid() == period.periodId() }
+            .takeIf { it.isNotEmpty() }
+            ?.let { matchingPeriods ->
+                val currentDate = Date()
+                matchingPeriods.any { dataInputPeriod ->
+                    val openingDate = dataInputPeriod.openingDate()
+                    val closingDate = dataInputPeriod.closingDate()
+
+                    openingDate == null && closingDate == null ||
```
vgarciabnz: This line, couldn't it be removed? The `null` value is already considered in the lines below. It seems a shortcut for the case where both are null. Or maybe `(openingDate == null || currentDate.before(openingDate)) && (...).

It is just to make it easier to read

vgarciabnz (commented):
Maybe we could have a unit test in DataSetInstanceServiceShould?

vgarciabnz (approved):

PR #2534, opened 2026-01-21
fix: [ANDROSDK-2149] Refactor default CategoryCombo handling to use dynamic server-provided UID

marcamsn:
The previous hardcoded UID assumed a specific DHIS2 server configuration. This change allows the SDK to work correctly with any DHIS2 instance regardless of its default CategoryCombo UID.

- Remove hardcoded `CategoryCombo.DEFAULT_UID` constant
- Add `DefaultCategoryComboManager` to dynamically fetch and cache the default CategoryCombo from the server
- Integrate default CategoryCombo fetch into `MetadataCall` during metadata sync
- Makes `categoryCombo` field non-nullable in `Program`, `DataElement`, and `DataSet` domain and db models
- Creates `CategoryComboWithFallbackDTO` to handle automatic fallback to default CategoryCombo during JSON deserialization
- Clear manager cache on logout


core/src/main/java/org/hisp/dhis/android/core/category/internal/DefaultCategoryComboManager.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private suspend fun loadDefaultsFromDatabase() {
+        if (defaultCategoryComboUid != null) return
+
+        val whereClause = WhereClauseBuilder()
+            .appendKeyNumberValue(CategoryComboTableInfo.Columns.IS_DEFAULT, 1)
+            .build()
+
+        categoryComboStore.selectOneWhere(whereClause)?.let { setDefaults(it) }
```
vgarciabnz: This object comes from the DB and does not include the nested `categoryOptionCombos` and `categories`, which are used in the `setDefaults()` method.

core/src/main/java/org/hisp/dhis/android/core/domain/metadata/MetadataCall.kt
```diff
@@ -148,6 +151,9 @@ internal class MetadataCall(
         constantModuleDownloader.downloadMetadata()
         emit(progressManager.increaseProgress(Constant::class.java, false))
 
+        defaultCategoryComboManager.fetchDefaults()
```
vgarciabnz, line 154: If I got it right, this request gets the defaults (categoryCombo, categoryOptionCombo and category) from the server but it doesn't persist them, they are just hold in memory to fill the potential nulls coming from the server.

If so, could we avoid this request on consecutive metadata syncs? I mean, after the first sync, these values would exist in the database. We would avoid a request on every metadata sync.

vgarciabnz (commented):
There are three elements with categoryCombos, two of them hava a nonNull categoryCombo in db (dataset, dataElement), the other one has a nullable categoryCombo in db (program). What if we make them all nonNull at db level and at domain level? They will be nullable at network level and the default will be assigned when mapping to domain. In this way, we could work with nonNulls at domain level, which is safer.

core/src/main/java/org/hisp/dhis/android/core/category/internal/DefaultCategoryComboManager.kt
marcamsn: I will use the repository instead of the store to retrieve the categoryCombo with the coc and categories. :)

core/src/main/java/org/hisp/dhis/android/core/domain/metadata/MetadataCall.kt
marcamsn: We can do that. Initially, I did it that way because it seemed a bit risky to assume those values would never change. But if we assume that, we could avoid it if we prefer.

core/src/main/java/org/hisp/dhis/android/core/category/internal/DefaultCategoryComboManager.kt
```diff
@@ -0,0 +1,81 @@
...
+@Singleton
+internal class DefaultCategoryComboManager(
+    private val categoryComboStore: CategoryComboStore,
+    private val networkHandler: CategoryComboNetworkHandler,
+) {
+    private var defaultCategoryComboUid: String? = null
+    private var defaultCategoryOptionComboUid: String? = null
+    private var defaultCategoryUid: String? = null
+
+    fun getDefaultCategoryComboUid(): String? = defaultCategoryComboUid
```
taridepaco: These methods are running blocking functions but they are also being called by suspend functions. As they are internal, they can be kept suspend functions and only be called by blocking ones on the public methods, imo.

core/src/main/java/org/hisp/dhis/android/network/dataset/DataSetDTO.kt
```diff
@@ -104,7 +105,7 @@ internal data class DataSetDTO(
             workflow(workflow?.toDomain())
             dataSetElements(dataSetElements.map { it.toDomain() })
             indicators(indicators.map { Indicator.builder().uid(it.id).build() })
-            DataSetInternalAccessor.insertSections(this, sections.map { it.toDomain() })
+            DataSetInternalAccessor.insertSections(this, sections.map { it.toDomain(categoryCombo.toDomain()) })
```
vgarciabnz: This is the Dataset categoryCombo (attributeCombo), and it is passed to the sections, which then assigns this combo to the dataElements in the section. This is not right, they are different categoryCombos.

I think the problem is in the Section domain model. It has a list of DataElements, but the DTO does not download the full dataElement but an ObjectWithUid. The list of DataElements is only used in the collectionRepository to add the nested list of DataElements when the app uses the `withDataElements` method.

vgarciabnz (approved):

PR #2542, opened 2026-02-05
chore: [ANDROSDK-2228] Mark TRACKER_ASSOCIATE value type as deprecated in v43

taridepaco:
This PR adds a new annotation for marking classes or methods with a deprecated tag, with the server version as a parameter. 
Also, the annotation LegacyDataValueApi has been moved to the corresponding annotation folder.

Related task: [ANDROSDK-2228](https://dhis2.atlassian.net/browse/ANDROSDK-2228)

[ANDROSDK-2228]: https://dhis2.atlassian.net/browse/ANDROSDK-2228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/androidTest/java/org/hisp/dhis/android/testapp/datavalue/DataValueCollectionRepositoryMockIntegrationShould.kt
```diff
@@ -244,7 +243,6 @@ class DataValueCollectionRepositoryMockIntegrationShould : BaseMockIntegrationTe
     }
 
     @Test
-    @OptIn(LegacyDataValueApi::class)
```
vgarciabnz, line 247: Why is it no longer needed?

core/src/main/java/org/hisp/dhis/android/core/systeminfo/DHISVersion.kt
```diff
@@ -27,22 +27,38 @@
  */
 package org.hisp.dhis.android.core.systeminfo
 
+const val VERSION_2_29 = "2.29"
```
vgarciabnz: Shouldn't they be internal? This is the prefix for each version, I think it is not very useful for the app, at least not under this name. We could expose the `prefix` property in the enum if we want to make it accessible to the app.

vgarciabnz (approved):

PR #2546, opened 2026-02-06
feat: Add property `priority` to ProgramRuleAction model [ANDROSDK-2222]

fernandogmz:
* Update domain, network and db models to support 'priority' property on ProgramRuleAction
* Update Test files for Integration
* Update migration script to support 'priority' column


vgarciabnz:
It looks good! @fernandogmz 
There are two missing things:
- Add the property in the `ProgramRuleActionSamples` object. This object is used to test the ProgramRuleAction table: it stores an object, reads it and checks that it is the same as the original. It is a bit redundant because the table is already tested with other tests, but it is good to have a specific test for this.
- Run the command `./gradlew :core:apiDump` to update the `core/api` file. This file is a dump of the SDK public API. It is useful to avoid unintentional changes in the public API. Then, you have to add this file to the PR.

vgarciabnz:
The Instrumented test are failing because of an unauthorized request to Browserstack. I think the action has no access to the Browserstack token because it is an external repository. I'll take a look to see if there is any option to allow it, but don't worry, we can run the tests locally and ignore this check.

vgarciabnz:
The Code quality checks are failing because of Ktlint. You can run `./gradlew ktlintCheck` to know what is failing. Or `./gradlew clean ktlintFormat` to allow the linter to autofix the problem if possible.

vgarciabnz (approved):
Looks good to me!

PR #2549, opened 2026-02-11
feat: Add `legendSet` to ProgramRuleAction model [ANDROSDK-2234]

fernandogmz:
* Update domain, network and db models to support `legendSet` property on ProgramRuleAction
* Update Test files for Integration
* Update migration script to support `legendSet` column


vgarciabnz (commented):
Great, the PR is adding the property correctly including all the tests.
Only one thing is missing: the SDK does not download all the LegendSet objects by the default but those linked to metadata downloaded by the user. The class `LegendSetUidsSeeker` looks for all the LegendSet uids linked to metadata in the database so they can be downloaded. It is necessary to tell the class that it has to check the ProgramRuleAction table as well.
The tests are not failing because the legendSet assigned to the test programRuleAction is the same as for other metadata, so it is already downloaded and the FK is fulfilled.

vgarciabnz (approved):

PR #2550, opened 2026-02-12
fix: [ANDROSDK-2245] EventService returns event is not editable even if the ownership belongs to the user

taridepaco:
This PR solves checking if an event is editable based on the user's scope, now the real organisation unit ownership is check to determine this, instead of the standard org unit attribute.

Related task: [ANDROSDK-2245](https://dhis2.atlassian.net/browse/ANDROSDK-2245)

[ANDROSDK-2245]: https://dhis2.atlassian.net/browse/ANDROSDK-2245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/event/internal/EventServiceImpl.kt
```diff
@@ -158,6 +163,30 @@ internal class EventServiceImpl(
         return Single.just(blockingCanAddEventToEnrollment(enrollmentUid, programStageUid))
     }
 
+    private fun isInCaptureScope(event: Event): Boolean {
+        val ownerOrgUnits = runBlocking { getEventOwnerOrgUid(event) }
+        return ownerOrgUnits.any { organisationUnitService.blockingIsInCaptureScope(it) }
+    }
+
+    private suspend fun getEventOwnerOrgUid(event: Event): List<String> {
```
vgarciabnz: Shouldn't it return a String instead of a list of Strings? It can be nullable if needed

core/src/main/java/org/hisp/dhis/android/core/event/internal/EventServiceImpl.kt
```diff
@@ -122,7 +127,7 @@ internal class EventServiceImpl(
             event.enrollment()?.let { !enrollmentService.blockingIsOpen(it) } ?: false ->
                 EventEditableStatus.NonEditable(EventNonEditableReason.ENROLLMENT_IS_NOT_OPEN)
 
-            event.organisationUnit()?.let { !organisationUnitService.blockingIsInCaptureScope(it) } ?: false ->
+            !isInCaptureScope(event) ->
```
vgarciabnz: Maybe a different name, like `isOwnedByUser` or `isOwnerOrgunitInCaptureScope`.

core/src/main/java/org/hisp/dhis/android/core/event/internal/EventServiceImpl.kt
vgarciabnz: The primary key is <program, trackedEntityInstance>, so the query will return 1 query maximum.

vgarciabnz (approved):

PR #2553, opened 2026-02-18
feat: [ANDROSDK-2246] Download and persist enrollment AOC metadata

taridepaco:
This PR adds enrollmentCategoryCombo to programs as a mandatory uid field. The default category combo is used as a fallback for the migration and when fetching from the server.

Related task: [ANDROSDK-2246](https://dhis2.atlassian.net/browse/ANDROSDK-2246)

[ANDROSDK-2246]: https://dhis2.atlassian.net/browse/ANDROSDK-2246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/sharedTest/resources/program/programs.json
```diff
@@ -59,6 +59,9 @@
       "categoryCombo": {
         "id": "m2jTvAj5kkm"
       },
+      "enrollmentCategoryCombo": {
```
vgarciabnz: Maybe we might not include the enrollmentCategoryCombo in one of the programs to test if the default is correctly applied in the network layer

vgarciabnz (changes requested):
The class `ProgramFields` has not been updated with the new property. I assume you will notice it when testing against a real instance

vgarciabnz (approved):

PR #2555, opened 2026-02-23
feat: [ANDROSDK-2253] Support for attributeOptionCombo on the Enrollment model

marcamsn:
- Add an `attributeOptionCombo` field to the Enrollment model as a non-null property, with a database migration that defaults existing enrollments to the default `CategoryOptionCombo`.
- Download this field from `v2.43+` and fall back to the default COC for older servers that don’t provide it.
- Include the field in upload payloads (new Tracker only), support filtering via `byAttributeOptionComboUid()`, and expose a setter in `EnrollmentObjectRepository`.
- Support `attributeOptionCombo` in `EnrollmentCreateProjection` with automatic default assignment.


core/src/sharedTest/resources/enrollment/enrollment.json
```diff
@@ -13,6 +13,7 @@
   "completedDate": "2014-08-21T12:27:50.730",
   "status": "COMPLETED",
   "deleted": false,
+  "attributeOptionCombo": "HllvX50cXC0",
```
vgarciabnz: This property is not going to be returned in the old enrollment syntax, we could remove it from here.

core/src/sharedTest/resources/trackedentity/tracked_entity_instance.json
```diff
@@ -33,6 +33,7 @@
       "deleted": false,
       "incidentDate": "2017-12-12T07:33:52.993",
       "status": "ACTIVE",
+      "attributeOptionCombo": "HllvX50cXC0",
```
vgarciabnz: Enrollments in old TEI payload won't include this property

core/src/main/assets/migrations/180.sql
```diff
@@ -47,3 +47,9 @@ ALTER TABLE Program RENAME TO Program_Old;
...
 DROP TABLE IF EXISTS Program_Old;
+
```
vgarciabnz, line 65: Consider adding the default category if it is possible that dbs out there don't contain it (to be discuss separately).

core/src/main/java/org/hisp/dhis/android/core/enrollment/EnrollmentCreateProjection.java
```diff
@@ -44,6 +45,13 @@ public abstract class EnrollmentCreateProjection {
     @NonNull
     public abstract String trackedEntityInstance();
 
+    @Nullable
+    public abstract String attributeOptionCombo();
+
+    /**
+     * @deprecated replaced by {@link #create(String, String, String, String)}
```
vgarciabnz: Do we want to mark it as deprecated? Or just leave it as an alternative builder (in case you don't need to specify the aoc)?

vgarciabnz (commented):
The test `NewTrackerImporterEnrollmentShould` is not modified, it should be.

core/src/main/java/org/hisp/dhis/android/core/enrollment/EnrollmentCreateProjection.java
marcamsn: Okay, let's remove this deprecation.

vgarciabnz (approved):

PR #2564, opened 2026-03-06
chore: [ANDROSDK-2261] Migrate internal import models to Kotlin data classes 

taridepaco:
In this PR, several models in Java from the imports folder are migrated to Kotlin data classes. The classes that are using them have been updated as well to adapt to the new way of getting properties and instantiating objects.

Related task: [ANDROSDK-2261](https://dhis2.atlassian.net/browse/ANDROSDK-2261)

[ANDROSDK-2261]: https://dhis2.atlassian.net/browse/ANDROSDK-2261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/api/core.api
```diff
@@ -6097,287 +6097,33 @@ public final class org/hisp/dhis/android/core/imports/TrackerJobManager {
...
-	public abstract fun reference ()Ljava/lang/String;
-	public abstract fun responseType ()Ljava/lang/String;
-}
-
-public final class org/hisp/dhis/android/core/imports/internal/DataValueImportSummaryWebResponse : org/hisp/dhis/android/core/imports/internal/WebResponse {
-	public fun <init> (Lorg/hisp/dhis/android/core/imports/internal/DataValueImportSummary;Ljava/lang/String;ILjava/lang/String;Ljava/lang/String;)V
-	public final fun component1 ()Lorg/hisp/dhis/android/core/imports/internal/DataValueImportSummary;
-	public final fun component2 ()Ljava/lang/String;
-	public final fun component3 ()I
+public final class org/hisp/dhis/android/core/imports/internal/HttpMessageResponse : org/hisp/dhis/android/core/imports/internal/WebResponse {
```
vgarciabnz, line 6100: Could these classes be internal? Maybe not if there are used in the app or we want to expose them to other apps, but it looks like they could be internal.

core/src/main/java/org/hisp/dhis/android/network/common/dto/BaseImportSummaryDTO.kt
```diff
@@ -39,14 +36,3 @@ internal interface BaseImportSummaryDTO {
     val conflicts: List<ImportConflictDTO>?
     val description: String?
 }
-
-internal fun <T> T.applyImportSummaryFields(item: BaseImportSummaryDTO): T where
```
vgarciabnz, line 43: We will lose these shortcuts if we decide to remove the builders for the models at some point. Just something to take into account. I wouldn't keep the builders only because of this though.

core/api/core.api
taridepaco: This and HttpMessageResponse need to be public because they are in the end linked to UserModule

vgarciabnz (approved):

PR #2565, opened 2026-03-06
fix: [ANDROSDK-2258] Improve JaCoCo code coverage reporting and add generic Store tests

marcamsn:
The plan is to improve JaCoCo code coverage so we have:

- Add generic tests in abstract Store integration base classes
- Fix JaCoCo reporting 0% coverage for migration classes by preventing KSP from generating `RoomGeneratedMigrations` for test variants `debugAndroidTest`/`debugUnitTest`, which were shadowing the JaCoCo-instrumented library classes in the test APK
- Add `DatabaseSQLiteConnectionMigrationShould` test to cover the `migrate(SQLiteConnection)` overload, bringing migration coverage from ~58% to ~100%
- Improve JaCoCo source directories to include Kotlin sources and KSP-generated code
- Exclude Room-generated DAO inner classes (`*Dao_Impl$*`) from JaCoCo report (uncoverable boilerplate)

### DIFF
--- a/core/src/androidTest/java/org/hisp/dhis/android/core/data/database/ObjectStoreAbstractIntegrationShould.kt
+++ b/core/src/androidTest/java/org/hisp/dhis/android/core/data/database/ObjectStoreAbstractIntegrationShould.kt
@@ -31,6 +31,7 @@ import com.google.common.truth.Truth.assertThat
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.runTest
 import org.hisp.dhis.android.core.arch.db.access.DatabaseAdapter
+import org.hisp.dhis.android.core.arch.db.sqlorder.internal.SQLOrderType
 import org.hisp.dhis.android.core.arch.db.stores.internal.ObjectStore
 import org.hisp.dhis.android.core.arch.db.tableinfos.TableInfo
 import org.hisp.dhis.android.core.common.CoreObject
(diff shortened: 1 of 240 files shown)


core/src/androidTest/java/org/hisp/dhis/android/core/data/database/ObjectStoreAbstractIntegrationShould.kt
```diff
@@ -49,6 +50,7 @@ abstract class ObjectStoreAbstractIntegrationShould<M : CoreObject> internal con
     private val databaseAdapter: DatabaseAdapter
 
     protected abstract fun buildObject(): M
+    protected open fun buildObjectWithNullableFields(): M? = null
```
vgarciabnz, line 54: Why is it optional? Should it be better to make it mandatory (as `buildObject`) so that it is noticed at compile time when not provided? If a store test does not provide this method, it will simply end the test silently and successfully.

marcamsn: I made it optional because not every object has nullable fields.
But maybe we can change how it is used in the tests a bit so we don't mark the tests as pass, but as skipped:

```
// Before
val nullableObject = buildObjectWithNullableFields() ?: return@runTest
```

```
// After
val nullableObject = buildObjectWithNullableFields()
Assume.assumeNotNull(nullableObject)
nullableObject!!
```

core/src/test/java/org/hisp/dhis/android/core/category/CategoryShould.kt
```diff
@@ -35,6 +35,8 @@ import org.junit.Test
 
 class CategoryShould : CoreObjectShould("category/category.json") {
 
+    override fun roundTripSerializer() = CategoryDTO.serializer()
```
vgarciabnz: This roundTripSerializer, is it added in all the classes? Maybe it could be added as a property of `CoreObjectShould`.

Or at least it should be reused in rest of the tests (a few lines below this)

vgarciabnz (approved):

PR #2585, opened 2026-04-21
fix: [ANDROSDK-2284] enrollmentCategoryCombo is not downloaded in some cases

taridepaco:
This PR adds the field enrollmentCategoryCombo to the list of category combo uids fetched and downloaded, since previously due to the field not having the default "categoryCombo" column name, the associated CC were not downloaded and thus, the complete program and related data was removed by the Foreign Key Cleaner. Also, a test has been added that checks that all the combination of column and tables swept for CC uids are equal to the columns that reference the CC table using DB commands, as a safe guard to prevent this happening in the future.

Related task: [ANDROSDK-2284](https://dhis2.atlassian.net/browse/ANDROSDK-2284)

[ANDROSDK-2284]: https://dhis2.atlassian.net/browse/ANDROSDK-2284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/androidTest/java/org/hisp/dhis/android/core/category/internal/CategoryComboUidsSeekerMockIntegrationShould.kt
+++ b/core/src/androidTest/java/org/hisp/dhis/android/core/category/internal/CategoryComboUidsSeekerMockIntegrationShould.kt
@@ -28,9 +28,17 @@
 package org.hisp.dhis.android.core.category.internal
 
 import com.google.common.truth.Truth.assertThat
+import com.google.common.truth.Truth.assertWithMessage
 import kotlinx.coroutines.test.runTest
 import org.hisp.dhis.android.core.arch.d2.internal.DhisAndroidSdkKoinContext.koin
+import org.hisp.dhis.android.core.category.CategoryCombo
+import org.hisp.dhis.android.core.common.ObjectWithUid
+import org.hisp.dhis.android.core.program.Program
+import org.hisp.dhis.android.core.program.internal.ProgramStore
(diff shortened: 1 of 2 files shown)


core/src/androidTest/java/org/hisp/dhis/android/core/category/internal/CategoryComboUidsSeekerMockIntegrationShould.kt
```diff
@@ -44,4 +52,82 @@ class CategoryComboUidsSeekerMockIntegrationShould : BaseMockIntegrationTestFull
...
+            .map { it.name }
+            .filter { it !in sqliteInternalTables && !it.startsWith("sqlite_") }
+            .filter { it !in excludedTables }
+
+        val discoveredReferences = mutableSetOf<Pair<String, String>>()
+        for (table in tableNames) {
+            val fkRows = databaseAdapter.rawQueryWithTypedValues("PRAGMA foreign_key_list(`$table`);")
+            for (row in fkRows) {
+                val values = row.values.toList()
+                if (values.size >= 5 && values[2].toString() == categoryComboTable) {
```
taridepaco, line 116: Values.size >= 5  is a safe check that only allows PRAGMA results with enough results.
The value for index = 2 corresponds to the table name
The value for index = 3 corresponds to the "from" column name

vgarciabnz (approved):

PR #2587, opened 2026-04-22
chore: [ANDROSDK-2281] Use setup-gradle action; update dependencies

vgarciabnz:
Changes:
- Use the `setup-gradle` action, which has a better cache management compared to `setup-java`. 
- Add an initial checkout step. _(only in develop, master and *-rc)._
- Update action versions.


vgarciabnz:
I am thinking that the initial "checkout" job is useless in branches other than "develop" or "master". The whole point of this initial job was to cache the build, but the cache is only saved for "develop" and "master" branches. For the rest of branches, the cache is just read-only.

taridepaco (approved):

PR #2594, opened 2026-05-04
chore: [ANDROSDK-2289] Remove ObjectStyle.Builder class

vgarciabnz:
Removes the ObjectWithStyle.Builder inner interface from the ObjectWithStyle interface. This inner interface was used to type the builders and indicate that they include a `style(...)` builder property, which was used in the DTO -> domain and DB -> domain conversions. This inner interface introduced some complexity when it comes to inheritance, and it was even more complex with the approach of generated builders for data class. So it was better to just remove it.

This PR does:
- Change the `ObjectWithStyle` to a simple interface with the `style()` property.
- Add the `ObjectWithStyleKt` interface for the data classes.
- Create a standard `toDomain` method for DTO and DB conversions.
- Modify classes that extend `ObjectWithStyle` to remove the builder extension and add the builder property explicitly.


core/src/main/java/org/hisp/dhis/android/core/trackedentity/TrackedEntityAttribute.java
taridepaco, line 157: why did this model have the style builder setter already here?

vgarciabnz: It guess it was unintentional, it shouldn't have been there.

taridepaco (approved):

PR #2596, opened 2026-05-05
chore: [ANDROSDK-2288-DATAELEMENT] Migrate DataElement

vgarciabnz:
This model migration introduces the question about the ObjectStyle default value. In the legacy DataElement class, the default value was specified in the builder. It has been kept in the new builder as well. But it does not apply to the default constructor. We can consider if this is something that should go to the network layer.


core/src/main/java/org/hisp/dhis/android/core/fileresource/FileResourceRoutine.kt
```diff
@@ -67,7 +66,7 @@ internal class FileResourceRoutine(
     suspend fun internalDeleteOutdatedFileResources(after: Date? = null) {
         val dataElementsUids = dataElementCollectionRepository
             .byValueType().`in`(ValueType.FILE_RESOURCE, ValueType.IMAGE)
-            .suspendGet().map(DataElement::uid)
+            .suspendGet().map { it.uid }
```
vgarciabnz, line 69: This is required because of an ambiguity in the name. Now we have `uid()` and `uid`, and it is not able to differentiate the function reference.

taridepaco (approved):

taridepaco (approved):

PR #2600, opened 2026-05-11
chore: [ANDROSDK-2295] Migrate some independent classes in settings module

vgarciabnz:
This PR includes some independent classes in the Settings module. There are some changes in nullability, they are explained in the comments.


core/src/main/java/org/hisp/dhis/android/core/settings/internal/ApkDistributionVersion.kt
```diff
@@ -29,8 +29,8 @@
 package org.hisp.dhis.android.core.settings.internal
 
 internal data class ApkDistributionVersion(
-    val version: String?,
-    val downloadURL: String?,
+    val version: String,
```
vgarciabnz, line 32: These two fields can be considered notNull.

core/src/main/java/org/hisp/dhis/android/core/settings/GeneralSettings.kt
```diff
@@ -0,0 +1,86 @@
...
+
+import org.hisp.dhis.android.annotations.ModelBuilder
+import org.hisp.dhis.android.core.common.CoreObject
+import java.util.Date
+
+@ModelBuilder
+@Suppress("TooManyFunctions")
+data class GeneralSettings(
+    @Deprecated("Use SynchronizationSettings.dataSync() instead.")
+    internal val dataSync: DataSyncPeriod?,
```
vgarciabnz, line 39: There are two deprecated fields that could be hidden for the app (just keep the getters).
It is required to keep these fields in order to store the value that is returned in the getters.

core/src/main/java/org/hisp/dhis/android/core/settings/LatestAppVersion.kt
```diff
@@ -26,35 +26,25 @@
...
+package org.hisp.dhis.android.core.settings
 
-import androidx.annotation.Nullable;
+import org.hisp.dhis.android.annotations.ModelBuilder
+import org.hisp.dhis.android.core.common.CoreObject
 
-import com.google.auto.value.AutoValue;
+@ModelBuilder
+data class LatestAppVersion(
+    val version: String,
```
vgarciabnz, line 36: This property is not null in the DB and the DTO.

core/src/main/java/org/hisp/dhis/android/core/settings/SettingsAppInfo.kt
```diff
@@ -26,35 +26,11 @@
...
-    public abstract static class Builder {
-
-        public abstract Builder dataStoreVersion(SettingsAppDataStoreVersion dataStoreVersion);
-
-        public abstract Builder androidSettingsVersion(String androidSettingsVersion);
-
-        public abstract SettingsAppInfo build();
-    }
-}
+internal data class SettingsAppInfo(
```
vgarciabnz, line 33: This is internal, so the builder could be ignored. It was quite straightforward to replace the builder with the constructor.

taridepaco (approved):

PR #2603, opened 2026-05-12
chore: [ANDROSDK-2295] Migrate SynchronizationSettings and SystemSetting

vgarciabnz:
Migrates:
- SynchronizationSettings and all the related files.
- SystemSetting.


core/src/main/java/org/hisp/dhis/android/core/settings/SystemSetting.kt
```diff
@@ -26,20 +26,21 @@
...
+package org.hisp.dhis.android.core.settings
 
-import androidx.annotation.Nullable;
+import org.hisp.dhis.android.annotations.ModelBuilder
+import org.hisp.dhis.android.core.common.CoreObject
 
-import com.google.auto.value.AutoValue;
+@ModelBuilder
+data class SystemSetting(
+    val key: SystemSettingKey?,
```
taridepaco: could this be non null? it seems in the db and dto it is expected to be always there

vgarciabnz: yes, good catch, I will change it

taridepaco (approved):

PR #2604, opened 2026-05-13
chore: [ANDROSDK-2295] Migrate AppearanceSettings

vgarciabnz:
It migrates all the classes related to AppearanceSettings. There are some changes compared to the original classes, they are explained in the comments.


core/src/main/java/org/hisp/dhis/android/core/settings/AppearanceSettings.kt
```diff
@@ -26,36 +26,31 @@
...
 
-import java.util.Map;
+    fun filterSorting(): FilterSorting? = filterSorting
+    fun programConfiguration(): ProgramConfigurationSettings? = programConfiguration
+    fun dataSetConfiguration(): DataSetConfigurationSettings? = dataSetConfiguration
 
-@AutoValue
-public abstract class DataSetConfigurationSettings {
+    @Deprecated("Use programConfiguration instead")
+    fun completionSpinner(): CompletionSpinnerSetting = programToCompletionSpinner(programConfiguration)
```
vgarciabnz, line 46: This function does not have a backing property in the constructor. It is evaluated at runtime if requested. The DTO translates from the legacy `completionSpinner` (when returned) to `programConfiguration`. From there on, the `programConfiguration` is the only property.

core/src/main/java/org/hisp/dhis/android/core/settings/AppearanceSettingsHelper.kt
```diff
@@ -28,20 +28,16 @@
...
 package org.hisp.dhis.android.core.settings
 
-import org.hisp.dhis.android.core.common.ObjectWithUidInterface
-
 internal object AppearanceSettingsHelper {
 
-    @JvmStatic
-    fun <O : ObjectWithUidInterface> getGlobal(list: List<O>): O? {
-        return list.find { it.uid() == null }
+    fun <O> getGlobal(list: List<O>, getId: (O) -> String?): O? {
```
vgarciabnz, line 33: We had a problem here. Previously, the classes `DataSetConfigurationSetting` and `ProgramConfigurationSetting` extended `ObjectWithUidInterface`, which had a property `uid()` with undefined nullability. We have defined this property as nonNull in the interface, so the problem comes when those classes are translated to Kotlin. Their `uid()` cannot be nonNull because it is null for the "global" configuration.

We had two options:
- Just remove the interface and tell the method a way to get the uid (this is the implemented option).
- Create a `ObjectWithNullableUid` interface and use it for those classes.

core/src/main/java/org/hisp/dhis/android/core/settings/AppearanceSettingsObjectRepository.kt
```diff
@@ -69,20 +68,19 @@ class AppearanceSettingsObjectRepository internal constructor(
 
             // ProgramConfigurationSettings
             val programConfigurationSettings = ProgramConfigurationSettings.builder()
-                .globalSettings(getGlobal(programConfigurationSettingList))
-                .specificSettings(getSpecifics(programConfigurationSettingList))
+                .globalSettings(getGlobal(programConfigurationSettingList) { it.uid })
```
vgarciabnz, line 71: Removing the `ObjectWithUidInterface` from `DataSetConfigurationSetting` and `ProgramConfigurationSetting` makes us write this extra code all along this class.

taridepaco (approved):

PR #2605, opened 2026-05-14
chore: [ANDROSDK-2292] allow multiaccount and add PIN to Oauth2

taridepaco:


core/src/main/java/org/hisp/dhis/android/core/user/oauth2/OAuth2Handler.kt
```diff
@@ -47,6 +47,10 @@ interface OAuth2Handler {
 
     fun getClientId(): String?
 
+    fun setPin(pin: String): Result<Unit>
```
vgarciabnz, line 50: This is kotlin native `Result` class, isn't it? We don't usually return this class but our own `Result` implementation (defined in the SDK), which has the option to define the failure type. 
I would change it just for consistency with the rest of the SDK. We can review it.

core/src/main/java/org/hisp/dhis/android/core/user/oauth2/internal/OAuth2StateSecureStore.kt
```diff
@@ -25,41 +25,46 @@
...
-    class Builder : DatabaseAccountBuilder()
+    fun remove(serverUrl: String, username: String) {
+        secureStore.removeData(buildKey(serverUrl, username))
+    }
+
+    private fun buildKey(serverUrl: String, username: String): String {
+        val normalized = ServerUrlNormalizer.normalize(serverUrl)
+        val digest = MessageDigest.getInstance(HASH_ALGORITHM)
+            .digest("$normalized|$username".toByteArray())
+        val hash = digest.take(HASH_BYTE_COUNT).joinToString("") {
```
vgarciabnz, line 57: Does it mean that it only takes 4 chars from the hash? Maybe it could be a bit longer to avoid clashes...

vgarciabnz (approved):

PR #2606, opened 2026-05-15
chore: [ANDROSDK-2298] migrate java enums to kotlin

taridepaco:
Related task: [ANDROSDK-2298](https://dhis2.atlassian.net/browse/ANDROSDK-2298)

[ANDROSDK-2298]: https://dhis2.atlassian.net/browse/ANDROSDK-2298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/settings/DownloadPeriod.kt
```diff
@@ -25,33 +25,31 @@
...
+    ;
 
-    DownloadPeriod(final Integer months) {
-        this.months = months;
-    }
+    val months: Int?
 
-    public Integer getMonths() {
-        return this.months;
+    init {
```
vgarciabnz: I think we don't need this `init` block. The property `months` can be exposed in the constructor as a "val".

core/src/main/java/org/hisp/dhis/android/core/settings/DownloadPeriod.kt
```diff
@@ -25,33 +25,31 @@
...
+        this.months = months
     }
 
-    public static DownloadPeriod forName(String name) {
-        for (DownloadPeriod c : values()) {
-            if (c.name().equals(name)) {
-                return c;
+    companion object {
+        fun forName(name: String?): DownloadPeriod? {
+            for (c in entries) {
```
vgarciabnz: This block could be replaced by a `find` function, which returns a null if not found. The same applies to other enums in this PR.

core/src/main/java/org/hisp/dhis/android/core/sms/domain/repository/internal/SubmissionType.kt
```diff
@@ -25,24 +25,20 @@
...
+    ;
 
-    SubmissionType(String text) {
-        this.text = text;
-    }
+    val text: String?
 
-    public String getText() {
-        return text;
+    init {
```
vgarciabnz: This `init` block could be removed, just make the `text` property public in the constructor.

core/src/main/java/org/hisp/dhis/android/core/validation/ValidationRuleOperator.kt
```diff
@@ -0,0 +1,46 @@
...
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.validation
+
+enum class ValidationRuleOperator(val mathematicalOperator: String, val apiName: String) {
```
vgarciabnz, line 30: This is changing the public API, this enum is part of ValidationRule. Is is required to make it uppercase?

vgarciabnz (approved):

PR #2609, opened 2026-05-20
chore: [ANDROSDK-2295] Migrate AnalyticsSettings and related classes

vgarciabnz:
Migrate AnalyticsSettings and the related classes to a data class.

This PR makes some changes in the nullability of some properties. There are two main cases:
- properties that were nonNull in the DTO and were nullable in domain and DB. They were nullable because of the link properties, which were defined in the handler. The definition of the link has been moved to the DTO, so they can be nonNull in domain and DB. There is a migration in the DB to change the table structure.
- properties that were nonNull in the DTO and DB, but were nullable in domain. The reason was the same, the link properties. There were some `!!` in the DB mapping. Moving the definition of the link to the DTO allowed making it nonNull in domain as well.


core/src/main/java/org/hisp/dhis/android/core/settings/AnalyticsTeiData.kt
```diff
@@ -0,0 +1,55 @@
...
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.settings
+
+import org.hisp.dhis.android.annotations.ModelBuilder
+
+@ModelBuilder
+data class AnalyticsTeiData(
+    val dataElements: List<AnalyticsTeiDataElement>,
```
taridepaco: could an emptyList() be used as default for the parameters in the constructor so the builder behavior is kept if we remove it?

taridepaco (dismissed):

PR #2610, opened 2026-05-20
chore: [ANDROSDK-2296] fetch oauth config json and persist device enrollment endpoints

taridepaco:
Downloading this file and temporarily store it will have two purposes:

1. determine if the server has oauth enabled, answering the app when requested this.
2. knowing the device enrollment endpoints instead of having those hardcoded

isOauthEnabled parameter of LoginConfig should be resolved based on the existence of this “well known json”

Related task [ANDROSDK-2296](https://dhis2.atlassian.net/browse/ANDROSDK-2296)

[ANDROSDK-2296]: https://dhis2.atlassian.net/browse/ANDROSDK-2296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/server/LoginConfig.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/server/LoginConfig.kt
@@ -49,14 +49,9 @@ data class LoginConfig(
     val useCustomLogoFront: Boolean = false,
     val oidcProviders: List<LoginOidcProvider> = emptyList(),
 ) {
-    fun isOauthEnabled(): Boolean {
-        return OAUTH_ENABLED
-    }
+    var isOauthEnabled: Boolean = false
 
     internal companion object {
-        // return false for SDK release 1.14.0 to prevent issues until fully implemented
-        private const val OAUTH_ENABLED = false
-
         fun createDefault(serverUrl: String): LoginConfig {
             return LoginConfig(
                 applicationTitle = serverUrl,
--- a/core/src/main/java/org/hisp/dhis/android/core/server/LoginConfig.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/server/LoginConfig.kt
@@ -50,6 +50,7 @@ data class LoginConfig(
     val oidcProviders: List<LoginOidcProvider> = emptyList(),
 ) {
     var isOauthEnabled: Boolean = false
+        internal set
 
     internal companion object {
         fun createDefault(serverUrl: String): LoginConfig {
--- a/core/src/main/java/org/hisp/dhis/android/core/server/OauthConfig.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/server/OauthConfig.kt
@@ -25,45 +25,9 @@
  *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package org.hisp.dhis.android.network.loginconfig
+package org.hisp.dhis.android.core.server
 
-import org.hisp.dhis.android.core.arch.api.HttpServiceClient
-import org.hisp.dhis.android.core.arch.api.executors.internal.CoroutineAPICallExecutor
-import org.hisp.dhis.android.core.server.LoginConfig
-import org.hisp.dhis.android.core.server.OauthConfig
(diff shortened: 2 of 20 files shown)


core/src/main/java/org/hisp/dhis/android/core/server/OauthConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.android.core.server
+
+data class OauthConfig(
```
vgarciabnz: Could it be internal?

core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
```diff
@@ -52,7 +54,9 @@ internal class OAuth2NetworkHandlerImpl(
         codeChallenge: String,
         scope: String,
     ): String {
-        return Uri.parse("$serverUrl/oauth2/authorize").buildUpon()
+        val authorizationEndpoint = oAuth2SecureStore.authorizationEndpoint
+            ?: error("Authorization endpoint not configured. checkServerUrl must run before OAuth login.")
+        return Uri.parse(authorizationEndpoint).buildUpon()
             .appendQueryParameter("client_id", clientId)
             .appendQueryParameter("redirect_uri", OAuth2Config.DEFAULT_REDIRECT_URI)
             .appendQueryParameter("response_type", "code")
```
vgarciabnz: There is a "response_type_methods" array in the .well-known. Could we use this value to avoid this hardcoded string? Maybe we don't actually support other methods and it has to be "code", I don't know.

core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
vgarciabnz: This information comes in the .well-known, there is a list of supported challenge methods. I don't think we should support any method defined in the file, but maybe we could validate if S256 is listed as a supported method. 

We can create different tasks to improve the support and the error messages if, for example, this method is not supported.

core/src/main/java/org/hisp/dhis/android/core/server/LoginConfig.kt
```diff
@@ -49,14 +49,9 @@ data class LoginConfig(
     val useCustomLogoFront: Boolean = false,
     val oidcProviders: List<LoginOidcProvider> = emptyList(),
 ) {
-    fun isOauthEnabled(): Boolean {
-        return OAUTH_ENABLED
-    }
+    var isOauthEnabled: Boolean = false
```
vgarciabnz, line 52: Maybe we can add the line "internal set" just below to restrict the visibility of the setter. Just to make it more immutable from the outside.

vgarciabnz (commented):
The comments about the .well-know properties can be addressed in a different PR. Maybe we can just validate the content of .well-know and verify that if has support for the functions we use:
- code challenge method
- grant types (authorization_code, refresh_token)
- response types

core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
vgarciabnz: We could replace the hardcoded strings here (and the grant types and the response_type) by the variables defined in the OauthConfig object.

vgarciabnz (dismissed):

vgarciabnz (approved):

PR #2611, opened 2026-05-20
chore: [ANDROSDK-2295] Migrate CustomIntents and related classes

vgarciabnz:
Last batch for data classes in the settings folder. It migrates the CustomIntents class and the related ones.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentAttribute.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentAttribute.kt
@@ -30,13 +30,13 @@ package org.hisp.dhis.android.core.settings
 
 import org.hisp.dhis.android.annotations.ModelBuilder
 import org.hisp.dhis.android.core.common.CoreObject
+import org.hisp.dhis.android.core.common.ObjectWithUidInterfaceKt
 
 @ModelBuilder
 data class CustomIntentAttribute(
-    val uid: String,
+    override val uid: String,
     val customIntentUid: String,
-) : CoreObject {
-    fun uid(): String = uid
+) : CoreObject, ObjectWithUidInterfaceKt {
     fun customIntentUid(): String = customIntentUid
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentAttribute.kt
```diff
@@ -26,28 +26,25 @@
...
+package org.hisp.dhis.android.core.settings
 
-import androidx.annotation.Nullable;
+import org.hisp.dhis.android.annotations.ModelBuilder
+import org.hisp.dhis.android.core.common.CoreObject
 
-import com.google.auto.value.AutoValue;
+@ModelBuilder
+data class CustomIntentAttribute(
+    val uid: String,
```
taridepaco: Couldn't the uid interface be added here?

core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentDataElement.kt
```diff
@@ -26,27 +26,25 @@
...
 
-import com.google.auto.value.AutoValue;
+import org.hisp.dhis.android.annotations.ModelBuilder
+import org.hisp.dhis.android.core.common.CoreObject
 
-@AutoValue
-public abstract class CustomIntentRequestArgument {
+@ModelBuilder
+data class CustomIntentDataElement(
+    val uid: String,
```
taridepaco: same of uid interface here?

core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentAttribute.kt
vgarciabnz: Done

core/src/main/java/org/hisp/dhis/android/core/settings/CustomIntentDataElement.kt
vgarciabnz: Done

taridepaco (approved):

PR #2612, opened 2026-05-21
chore: [ANDROSDK-2296] add authorization type to databaseAccount

andresmr:
Add AuthorizationType property to DatabaseAccount model to identify the auth type: BASIC, OPEN_ID_CONNECT, OAUTH2


taridepaco (dismissed):

core/src/main/java/org/hisp/dhis/android/core/user/internal/LogInCall.kt
```diff
@@ -151,7 +151,11 @@ internal class LogInCall(
         credentialsSecureStore.set(credentials)
         userIdStore.set(user.uid())
 
-        loginDatabaseManager.loadDatabaseOnline(credentials.serverUrl, credentials.username)
+        loginDatabaseManager.loadDatabaseOnline(
+            credentials.serverUrl,
+            credentials.username,
+            credentialsSecureStore.getAuthorizationType(),
```
vgarciabnz: Is it not a bit weird to get the authorizationType from the credentialsStore if you already have the credentials object here?

taridepaco: The auth. type needs to be calculated from different credential parameters and it's used in different places, so isn't there the right place to have the logic? How can this be done differently?

PR #2613, opened 2026-05-22
feature: [ANDROSDK-1995] evaluate PI disaggregations in analytics

taridepaco:
Adapt the analytics service to evaluate Program Indicator (PI) disaggregations. The service now resolves disaggregation filters by querying the program filters store and builds the corresponding SQL `WHERE` clauses, with new unit tests and updated mock server fixtures covering the flow.

Related task: [ANDROSDK-1995](https://dhis2.atlassian.net/browse/ANDROSDK-1995)

[ANDROSDK-1995]: https://dhis2.atlassian.net/browse/ANDROSDK-1995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz (commented):
Good PR, short and clean!

There is another case that might not be properly managed. If the evaluationItem contains a Category of type "ATTRIBUTE", this Category will be considered in the disaggregation method. As the category uid won't be found in the PI categoryMappings, it will return a "0", which makes the "where" condition never succeed. It could be changed in another PR (if we decide to move it to the context) if preferred.

vgarciabnz (dismissed):

vgarciabnz (approved):

PR #2617, opened 2026-05-25
chore: [ANDROSDK-2307] migrate user module models to Kotlin

taridepaco:
Port the 7 AutoValue domain models in the user package to Kotlin data classes via the `@ModelBuilder` KSP annotation: `AuthenticatedUser`, `Authority`, `User`, `UserCredentials`, `UserGroup`, `UserOrganisationUnitLink` and `UserRole`. Public builders and accessor methods are preserved so existing Java and Kotlin call sites keep compiling.

Tightens `UserInternalAccessor.insertOrganisationUnits` to accept `List<OrganisationUnit>?` and updates `UserOrganisationUnitDB.toDomain()` to safely guard the now non-null `root` / `userAssigned` setters. The public API dump is regenerated accordingly.

Related task: [ANDROSDK-2307](https://dhis2.atlassian.net/browse/ANDROSDK-2307)

[ANDROSDK-2307]: https://dhis2.atlassian.net/browse/ANDROSDK-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/user/UserInternalAccessor.kt
vgarciabnz: Couldn't we remove this internalAccessor? Maybe it implies the refactor of some classes

core/src/main/java/org/hisp/dhis/android/persistence/user/UserOrganisationUnitDB.kt
```diff
@@ -28,13 +28,13 @@ internal data class UserOrganisationUnitDB(
...
-            .organisationUnitScope(organisationUnitScope)
-            .root(root)
-            .userAssigned(userAssigned)
-            .build()
+        return UserOrganisationUnitLink.builder().apply {
+            user(user)
+            organisationUnit(organisationUnit)
+            organisationUnitScope(organisationUnitScope)
+            root?.let { root(it) }
+            userAssigned?.let { userAssigned(it) }
```
vgarciabnz: This is risky because `root` and `userAssigned` are nonNull in domain, so it would fail when building the domain object if any of them are null in DB because they were not provided.
I think it is safe to make them nonNull in DB as well because it would be already failing in the AutoValue class if any of them were null in DB.

vgarciabnz (approved):

PR #2618, opened 2026-05-26
chore: [ANDROSDK-2308] migrate category models to kotlin

taridepaco:
Migrate the Category domain models from Java AutoValue classes to Kotlin data classes using the `@ModelBuilder` KSP annotation. The migration covers `Category`, `CategoryCombo`, `CategoryOptionCombo`, `CategoryCategoryComboLink`, `CategoryCategoryOptionLink`, `CategoryOptionComboCategoryOptionLink` and `CategoryOptionOrganisationUnitLink`, keeping the existing public builder API and field nullability so callers stay source-compatible.

Related task: [ANDROSDK-2308](https://dhis2.atlassian.net/browse/ANDROSDK-2308)

[ANDROSDK-2308]: https://dhis2.atlassian.net/browse/ANDROSDK-2308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/category/CategoryCombo.kt
```diff
@@ -26,27 +26,33 @@
...
     override val uid: String,
     override val code: String?,
     override val name: String?,
     override val displayName: String?,
     override val created: Date?,
     override val lastUpdated: Date?,
     override val deleted: Boolean?,
+    val isDefault: Boolean?,
+    val categories: List<Category>?,
+    internal val categoryOptionCombos: List<CategoryOptionCombo>?,
```
vgarciabnz, line 47: Is there an internal accessor we could remove?

core/src/main/java/org/hisp/dhis/android/persistence/category/CategoryOptionOrganisationUnitLinkDB.kt
```diff
@@ -41,7 +41,7 @@ internal data class CategoryOptionOrganisationUnitLinkDB(
         return CategoryOptionOrganisationUnitLink.builder()
             .categoryOption(categoryOption)
             .organisationUnit(organisationUnit)
-            .restriction(restriction)
+            .restriction(restriction!!)
```
vgarciabnz: If restriction was nonNull in domain before, we could make it nonNull in DB as well.

vgarciabnz (approved):

PR #2624, opened 2026-05-29
chore: [ANDROSDK-2311] MIgrate Relationship models to kotlin

taridepaco:
Relationhip models migrated, DeletableDataObjectKt now implements DeletableDataObject to prevent mismatch interfaces in the stores.

Related task: [ANDROSDK-2311](https://dhis2.atlassian.net/browse/ANDROSDK-2311)

[ANDROSDK-2311]: https://dhis2.atlassian.net/browse/ANDROSDK-2311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/relationship/Relationship.kt
```diff
@@ -0,0 +1,70 @@
...
+    val uid: String,
+    val name: String?,
+    val created: Date?,
+    val lastUpdated: Date?,
+    val syncState: State?,
+    override val deleted: Boolean?,
+    val relationshipType: String?,
+    val from: RelationshipItem?,
+    val to: RelationshipItem?,
+) : DeletableDataObjectKt, ObjectWithUidInterface {
```
vgarciabnz: There is an ObjectWithUidInterfaceKt that adds the `val uid`

core/src/main/java/org/hisp/dhis/android/core/relationship/RelationshipCollectionRepository.kt
```diff
@@ -105,14 +104,9 @@ class RelationshipCollectionRepository internal constructor(
                 .build()
         } else {
             val from = o.from()
-            relationshipWithUid = if (o.uid() == null) {
-                val generatedUid = UidGeneratorImpl().generate()
-                o.toBuilder().uid(generatedUid).build()
-            } else {
-                o
-            }
+            relationshipWithUid = o
```
vgarciabnz: This variable seems a bit redundant

core/src/main/java/org/hisp/dhis/android/network/relationship/RelationshipDTO.kt
```diff
@@ -49,7 +49,7 @@ internal data class RelationshipDTO(
 ) : BaseDeletableDataObjectDTO {
     fun toDomain(): Relationship {
         return Relationship.builder()
-            .uid(relationship)
+            .uid(relationship!!)
```
vgarciabnz: Maybe it makes more sense to make it nonNull in the DTO. It will fail anayway and seems clearer. The relationship field was added in 2.30, so it should be safe to mark it as nonNull.

core/src/main/java/org/hisp/dhis/android/network/tracker/NewRelationshipDTO.kt
```diff
@@ -51,7 +51,7 @@ internal data class NewRelationshipDTO(
 
     fun toDomain(): Relationship {
         return Relationship.builder()
-            .uid(relationship)
+            .uid(relationship!!)
```
vgarciabnz: Same here, mark the property as nonNull in the DTO

vgarciabnz (approved):

PR #2626, opened 2026-06-02
chore: [ANDROSDK-2317] migrate option domain models to kotlin

taridepaco:
Migrates the four AutoValue domain models in the `option` package (`Option`, `OptionGroup`, `OptionSet`, `OptionGroupOptionLink`) from Java to Kotlin data classes using the `@ModelBuilder` annotation, replacing the `BaseIdentifiableObAuVa` base with `BaseIdentifiableObjectKt`. This is part of the ongoing batch effort to migrate domain models to Kotlin and carries no user-visible behavior change.

Defaults previously applied in the AutoValue builders were preserved (e.g. the `ObjectStyle` default for `Option` now lives in its `builder()` factory), Java-style accessor methods were kept for compatibility, and a few now-non-null call sites were cleaned up. The public API dump and affected tests were updated accordingly.

Related task: [ANDROSDK-2317](https://dhis2.atlassian.net/browse/ANDROSDK-2317)

[ANDROSDK-2317]: https://dhis2.atlassian.net/browse/ANDROSDK-2317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/option/OptionGroup.kt
```diff
@@ -26,36 +26,33 @@
...
     override val name: String?,
     override val displayName: String?,
     override val created: Date?,
     override val lastUpdated: Date?,
     override val deleted: Boolean?,
-    val program: String?,
-    val programStage: String?,
-    val description: String?,
-    val eventQueryCriteria: EventQueryCriteria?,
+    val optionSet: ObjectWithUid?,
```
vgarciabnz: optionSet is nonNull in DB. It could be made nonNull in DTO and domain as well. There is a `!!` in the DB conversion, so it would fail already.

vgarciabnz (approved):

PR #2628, opened 2026-06-08
chore: [ANDROSDK-2315] Implement multi account and offline for OpenId connections

taridepaco:
Similarly to what was implemented for OAuth accounts, the OpenId Auth state and a PIN is persisted so several accounts with this configuration can be created and restored, allowing offline connections as well.

Related task: [ANDROSDK-2315](https://dhis2.atlassian.net/browse/ANDROSDK-2315)

[ANDROSDK-2315]: https://dhis2.atlassian.net/browse/ANDROSDK-2315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -19503,11 +19503,20 @@ public final class org/hisp/dhis/android/core/user/openid/OpenIDConnectConfig {
 }
 
 public abstract interface class org/hisp/dhis/android/core/user/openid/OpenIDConnectHandler {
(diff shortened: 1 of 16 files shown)


andresmr (approved):

vgarciabnz (approved):
Good one! Let's test it with app.
I wonder if the case of an existing openId account in the old format will be correctly managed, I am not sure if the store is completely new or it reuses the previous one somehow. We can create a ticket to check if the existing openId account keeps working with the app update. If not, we could log out the user and force them to login again using the new store.

PR #2629, opened 2026-06-08
[ANDROSDK-2319] fix: avoid re-downloading exhausted programs in tracker download

taridepaco:
## Summary

Fixes redundant tracked-entity downloads during a global sync. For a single program with few TEIs, the SDK was issuing the same paginated request several times across bundle iterations, re-downloading and re-persisting already-fetched data. The number of redundant calls grew quadratically with the number of programs in the bundle.

## Root cause

In `TrackerDownloadCall.iterateBundleOrgunit`, when a program was exhausted it was removed by filtering the **original captured `bundlePrograms` list** instead of the current value in `bundleOrgUnitPrograms`. Each exhaustion overwrote the previous assignment, so only the **last** exhausted program was actually removed per iteration. With M programs sharing the same org units (the global download case), this forced one extra bundle iteration per program and re-issued the same network calls — `M(M+1)/2` requests instead of `M`.

## Changes

1. **Main fix:** the removal of exhausted programs now operates cumulatively on the remaining programs, so each org-unit/program combination is queried only once when it is exhausted.
2. **Safety net:** track the number of items downloaded per iteration and stop the bundle loop as soon as a full iteration downloads nothing new, so termination no longer depends solely on the org-unit list being emptied.

## Testing

- Added a regression test (`TrackerDownloadCallShould`) that sets up a global bundle with multiple programs (each exhausted in its first page) and asserts each program is requested exactly once. The test fails against the previous logic (`TooManyActualInvocations`) and passes with the fix.
- `./gradlew :core:testDebugUnitTest --tests *TrackerDownloadCallShould` → green.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/tracker/exporter/TrackerDownloadCall.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/tracker/exporter/TrackerDownloadCall.kt
@@ -194,6 +194,7 @@ internal abstract class TrackerDownloadCall<T, Q : BaseTrackerQueryBundle>(
         val iterationResult = IterationResult()
 
         bundleResult.bundleOrgUnitPrograms[orgUnitUid]?.let { bundlePrograms ->
+            var remainingPrograms = bundlePrograms
             for (bundleProgram in bundlePrograms) {
                 if (bundleResult.bundleCount < bundle.commonParams().limit) {
                     val trackerQuery = getQuery(bundle, bundleProgram.program, orgUnitUid, limit)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/hisp/dhis/android/core/tracker/exporter/TrackerDownloadCall.kt
```diff
@@ -120,7 +120,10 @@ internal abstract class TrackerDownloadCall<T, Q : BaseTrackerQueryBundle>(
                     val result = iterateBundle(bundle, params, bundleResult, relatives, progressManager)
                     successfulSync = successfulSync && result.successfulSync
                     iterationCount++
-                } while (iterationNotFinished(bundle, params, bundleResult, iterationCount))
+                } while (
+                    result.downloadedCount > 0 &&
```
vgarciabnz, line 124: I think it makes sense to have this condition here. It doesn't save any API request but a loop iteration (just logic) to clean up the orgunit list.

vgarciabnz (approved):

PR #2630, opened 2026-06-08
fix: [ANDROAPP-7646]   fix: discard online failures in OFFLINE_FIRST to preserve local results

xurxodev:
Link to [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7646).

## Problem
  
In `OFFLINE_FIRST` mode, `TrackedEntityInstanceQueryDataFetcher` runs an online query to top up results when the offline page is short. If that online call fails (e.g. HTTP 400, network timeout), the failure item was`LoadResult.Error`. 

Paging3 then discards the offline results already collected for that page and the UI freezes on the previous result set — even though valid local data was available.
  
## Solution
  
Filter out failed (`Result.Failure`) online items before merging them with the local results, so a failing online query no longer wipes out valid offline data — the offline results are still emitted and the UI stays responsive. Local failures originating from `transform()` (e.g. missing `TrackedEntityType`) are unaffected, since they're produced after this merge point and still propagate as before.

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryDataFetcher.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryDataFetcher.kt
@@ -81,7 +81,7 @@ internal class TrackedEntityInstanceQueryDataFetcher(
             }
             if (result.size < requestedLoadSize && scope.mode() == RepositoryMode.OFFLINE_FIRST) {
                 val onlineInstances = queryOnline(requestedLoadSize)
-                result.addAll(onlineInstances)
+                result.addAll(onlineInstances.filter { it.succeeded })
             }
         } else {
             val instances = queryOnline(requestedLoadSize)


vgarciabnz:
Hi @xurxodev, thanks for the PR 👏 .

We wanted to properly fix this issue by showing an error message at the end of the list to inform the user that something went wrong online, but there was no time for the last release (hopefully in the next one).

This proposed fix is much better than the current situation where the offline results were just discarded with no feedback to user. So I think we can include this PR in the patch version and leave a more complete fix for the next version (October).

Could you change the base branch of the PR from `develop` to `1.4.1-rc`?

xurxodev:
@vgarciabnz I started from development, do you think make sense to do other against 1.14.1-rc?

vgarciabnz:
@xurxodev yes please, change the base or open a different PR using `1.14.1-rc` as the base branch so we can include this fix in the next patch version. The `develop` branch is way ahead right now. And the branch `1.14.1-rc` will be eventually merged into `develop`.

xurxodev:
@vgarciabnz Ready https://github.com/dhis2/dhis2-android-sdk/pull/2633

taridepaco:
Changes were merged into the upcoming patch 1.14.1 , so this one can be closed.

PR #2632, opened 2026-06-09
chore: [ANDROSDK-2318] Migrate model classes in Dataset package

vgarciabnz:
Migrate model classes in the dataset package.


core/src/main/java/org/hisp/dhis/android/core/dataset/DataSetInstance.kt
```diff
@@ -0,0 +1,83 @@
...
+    val attributeOptionComboDisplayName: String,
+    val valueCount: Int,
+    val completed: Boolean,
+    val completionDate: Date?,
+    val completedBy: String?,
+    val lastUpdated: Date?,
+    val dataValueState: State?,
+    val completionState: State?,
+    val state: State?,
+) : CoreObject {
```
vgarciabnz, line 56: I kept the `CoreObject` class here, although this class doesn't represent an actual DHIS2 class or a table in the DB, it is just a projection.

core/src/main/java/org/hisp/dhis/android/network/dataset/DataSetElementDTO.kt
```diff
@@ -29,19 +29,19 @@
...
 import kotlinx.serialization.Serializable
+import org.hisp.dhis.android.core.common.ObjectWithUid
 import org.hisp.dhis.android.core.dataset.DataSetElement
 import org.hisp.dhis.android.network.common.dto.ObjectWithUidDTO
 
 @Serializable
 internal data class DataSetElementDTO(
-    val dataSet: ObjectWithUidDTO?,
-    val dataElement: ObjectWithUidDTO?,
+    val dataElement: ObjectWithUidDTO,
```
vgarciabnz, line 38: The property dataset is removed and added in the `toDomain()` method.

core/src/main/java/org/hisp/dhis/android/network/dataset/SectionDTO.kt
```diff
@@ -50,22 +50,21 @@ internal data class SectionDTO(
     val sortOrder: Int?,
     val showRowTotals: Boolean?,
     val showColumnTotals: Boolean?,
-    val dataSet: ObjectWithUidDTO?,
```
vgarciabnz, line 53: This property is passed as a parameter in the `toDomain()` method.

taridepaco (approved):

PR #2639, opened 2026-06-16
chore: [ANDROSDK-2323] migrate enrollment domain models to kotlin

taridepaco:
Migrates the enrollment domain models (batch 10) from Java AutoValue classes to Kotlin data classes using the `@ModelBuilder` KSP annotation: `Enrollment` and `EnrollmentCreateProjection`. The public API and persistence behavior are kept unchanged, and the deprecated in-memory `coordinate` handling is dropped in favour of `geometry`.

As part of the migration, `EnrollmentInternalAccessor` is removed: the `events`/`relationships` properties become `internal` on the model and the call sites use them directly. `EnrollmentConverter` is ported to Kotlin to access those internal members, which required widening the abstract methods of the SMS `Converter` base (and its subclasses) to `protected` so a Kotlin subclass can override them. The API dump was regenerated accordingly.

Related task: [ANDROSDK-2323](https://dhis2.atlassian.net/browse/ANDROSDK-2323)

[ANDROSDK-2323]: https://dhis2.atlassian.net/browse/ANDROSDK-2323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/enrollment/EnrollmentCreateProjection.kt
```diff
@@ -0,0 +1,82 @@
...
+            return EnrollmentCreateProjection(
+                organisationUnit = organisationUnit,
+                program = program,
+                trackedEntityInstance = trackedEntityInstance,
+                attributeOptionCombo = null,
+            )
+        }
+
+        @JvmStatic
+        fun create(
```
vgarciabnz, line 68: Just as a comment, there is a function called [@JvmOverloads](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-overloads/) that gets the same behavior by using the default kotlin parameters, so we have just one function with the default value to `null`. It would be seen as two different functions from the java world.

vgarciabnz (approved):

PR #2641, opened 2026-06-17
chore: [ANDROSDK-2326] migrate Note, CustomIcon and FileResource to kotlin

taridepaco:
Migrates the `Note`, `NoteCreateProjection`, `CustomIcon` and `FileResource` domain models from Java AutoValue to Kotlin data classes using the `@ModelBuilder` annotation, continuing the ongoing model-to-Kotlin migration (batch 11). No user-visible behavior change.

For `FileResource`, `uid` is kept non-null (it is the database primary key), while the remaining fields stay nullable to preserve the previous AutoValue contract. `FileResourceStorageStatus` was made public because a public Kotlin data class cannot expose an `internal` type through its generated constructor/`copy()`; the public API dump was regenerated accordingly.

Related task: [ANDROSDK-2326](https://dhis2.atlassian.net/browse/ANDROSDK-2326)

[ANDROSDK-2326]: https://dhis2.atlassian.net/browse/ANDROSDK-2326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/fileresource/FileResourceCollectionRepository.kt
```diff
@@ -105,7 +105,7 @@ class FileResourceCollectionRepository internal constructor(
             val dstFile = saveFile(o, generatedUid, context)
             val fileResource = transformer.transform(dstFile).toBuilder().uid(generatedUid).name(o.name).build()
```
vgarciabnz, line 106: The `generatedUid` could be removed because it is assigned in the transformer, isn't it?

core/src/main/java/org/hisp/dhis/android/core/fileresource/FileResourceRoutine.kt
```diff
@@ -104,7 +104,7 @@ internal class FileResourceRoutine(
 
     private suspend fun deleteFileResources(fileResources: List<FileResource>) {
         fileResources.forEach { fileResource ->
-            fileResource.uid()?.let { uid ->
+            fileResource.uid.let { uid ->
```
andresmr: `uid` can't no longer be null, right?. If so, wouldn't be better remove the let for clarity?, something like:
`
fileResourceStore.deleteIfExists(fileResource.uid)
fileResource.path()?.let { path ->
   deleteFile(path)
}
`

vgarciabnz (approved):

PR #2642, opened 2026-06-17
chore: [ANDROSDK-2327] filter out category option mappings without filter

taridepaco:
Category option mappings are expected to always carry a filter, but some objects arrive from the server without one. Since the field was non-nullable, deserialization failed on those objects. This makes `CategoryOptionMappingDTO.filter` nullable and drops the option mappings without a filter when mapping a category mapping to its domain model, so the program download no longer fails on these entries.

A `program.json` test case with an option mapping lacking a filter was added to verify it is excluded after deserialization and `toDomain`.

Related task: [ANDROSDK-2327](https://dhis2.atlassian.net/browse/ANDROSDK-2327)

[ANDROSDK-2327]: https://dhis2.atlassian.net/browse/ANDROSDK-2327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/network/program/CategoryOptionMappingDTO.kt
+++ b/core/src/main/java/org/hisp/dhis/android/network/program/CategoryOptionMappingDTO.kt
@@ -34,13 +34,13 @@ import org.hisp.dhis.android.core.program.CategoryOptionMapping
 @Serializable
 internal data class CategoryOptionMappingDTO(
     val optionId: String,
-    val filter: String,
+    val filter: String? = null,
 ) {
     fun toDomain(categoryMappingId: String): CategoryOptionMapping {
         return CategoryOptionMapping.builder()
             .categoryMapping(categoryMappingId)
             .optionId(optionId)
-            .filter(filter)
+            .filter(filter!!)
             .build()
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/hisp/dhis/android/network/program/CategoryOptionMappingDTO.kt
```diff
@@ -34,13 +34,13 @@ import org.hisp.dhis.android.core.program.CategoryOptionMapping
...
     val optionId: String,
-    val filter: String,
+    val filter: String? = null,
 ) {
     fun toDomain(categoryMappingId: String): CategoryOptionMapping {
         return CategoryOptionMapping.builder()
             .categoryMapping(categoryMappingId)
             .optionId(optionId)
-            .filter(filter)
+            .filter(filter!!)
```
vgarciabnz: This double `!!` is a bit scary, although the list is filtered out in the other class and shouldn't be a problem. Couldn't we return a `null` if filter is not defined, and then use the `mapNotNull()` in the other class? I mean, just to avoid the `!!`.

vgarciabnz (approved):

PR #2645, opened 2026-06-25
chore: [ANDROSDK-2330] migrate map layer domain models to kotlin

taridepaco:
Migrate the map layer domain models (`MapLayer`, `MapLayerImageryProvider` and `MapLayerImageryProviderArea`) from Java AutoValue classes to Kotlin data classes using the `@ModelBuilder` annotation, as part of batch 12 of the domain model Kotlin migration.

As part of the migration the `MapLayer.external` property is tightened to non-null to match its API contract. This adds a migration in `181.sql` that recreates the `MapLayer` table with `external INTEGER NOT NULL` (dropping rows where it was null) and adapts `MapLayerDB` accordingly.

Related task: [ANDROSDK-2330](https://dhis2.atlassian.net/browse/ANDROSDK-2330)

[ANDROSDK-2330]: https://dhis2.atlassian.net/browse/ANDROSDK-2330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/map/layer/MapLayer.kt
```diff
@@ -26,42 +26,54 @@
...
+    fun layers(): String? = layers
+    fun linkedLayerUid(): String? = linkedLayerUid
 
-    class Builder : TrackerVisualizationDimensionBuilder()
+    fun toBuilder(): Builder = MapLayerBuilder.from(this)
+
+    class Builder : MapLayerBuilder()
 
     companion object {
+        const val IMAGERY_PROVIDERS = "imageryProviders"
```
vgarciabnz: I don't think this property was intentionally public in the model class. It seems it is only used for the childrenAppender identification, so it could be moved to the companion object in the collectionRepository as the rest of children appenders

vgarciabnz (dismissed):

vgarciabnz (approved):

PR #2648, opened 2026-06-26
chore: [ANDROSDK-2333] migrate conflict and data value models to kotlin

taridepaco:
Migrates the import/data-value conflict models and the `DataValue` model from Java AutoValue to Kotlin data classes using the `@ModelBuilder` annotation, continuing the model-migration effort (batch 15). `TrackerImportConflict` and `DataValueConflict` are straightforward `CoreObject` conversions. `DataValue` extends `BaseDeletableDataObject`, so it follows the `Event` precedent: it now implements `DataObjectKt`/`DeletableDataObjectKt` with `syncState`/`deleted` properties, and the builder's `SYNCED`/`false` defaults are preserved in the `builder()` factory to keep behavior identical across all construction paths.

Related task: [ANDROSDK-2333](https://dhis2.atlassian.net/browse/ANDROSDK-2333)

[ANDROSDK-2333]: https://dhis2.atlassian.net/browse/ANDROSDK-2333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/datavalue/DataValue.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.hisp.dhis.android.core.common.DeletableDataObjectKt
+import org.hisp.dhis.android.core.common.State
+import java.util.Date
+
+@ModelBuilder
+@Suppress("TooManyFunctions")
+data class DataValue(
+    override val syncState: State?,
+    override val deleted: Boolean?,
+    val dataElement: String?,
```
vgarciabnz: We could try to make these 5 properties non-null. They are non-null in the DTO and they are the primary key in the DB. Maybe it has more implications that I don't see now.

core/src/main/java/org/hisp/dhis/android/core/datavalue/DataValueConflict.kt
```diff
@@ -26,32 +26,44 @@
...
-    override val created: Date?,
-    override val lastUpdated: Date?,
-    override val deleted: Boolean?,
-    val number: Boolean?,
-    val factor: Int?,
-) : BaseIdentifiableObjectKt, CoreObject {
+data class DataValueConflict(
+    val conflict: String?,
+    val value: String?,
+    val attributeOptionCombo: String?,
```
vgarciabnz: Same here, these 5 properties are non-null in the db. This class is only generated internally in the SDK (it does not come from the API).

vgarciabnz (changes requested):

core/src/androidTest/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceCallRealIntegrationShould.kt
```diff
@@ -202,10 +202,10 @@ class FileResourceCallRealIntegrationShould : BaseRealIntegrationTest() {
...
                 nextPeriod.periodId()!!,
-                dataValue.organisationUnit()!!,
-                dataValue.dataElement()!!,
-                dataValue.categoryOptionCombo()!!,
-                dataValue.attributeOptionCombo()!!,
+                dataValue.organisationUnit(),
+                dataValue.dataElement(),
+                dataValue.categoryOptionCombo(),
+                dataValue.attributeOptionCombo(),
                 DataValueInternalAccessor.accessSourceDataSet(dataValue)!!,
```
vgarciabnz: This internal accessor, could it be removed?

vgarciabnz (approved):

PR #2650, opened 2026-06-29
chore: [ANDROSDK-2336] migrate use case and working list domain models to Kotlin

taridepaco:
Migrates the stock use case (`InternalStockUseCase`, `InternalStockUseCaseTransaction`) and program stage working list (`ProgramStageWorkingList`, `ProgramStageQueryCriteria`) domain models from Java AutoValue classes to Kotlin data classes annotated with `@ModelBuilder`, keeping the public API and behavior unchanged.

While aligning the models with their DTOs, several fields that were already non-null in the domain and DTO layers but nullable in the database were tightened to non-null in the persistence entities. A migration in `181.sql` removes rows holding null values for those columns and recreates the `StockUseCase` and `StockUseCaseTransaction` tables with the corresponding `NOT NULL` constraints.

Related task: [ANDROSDK-2336](https://dhis2.atlassian.net/browse/ANDROSDK-2336)

[ANDROSDK-2336]: https://dhis2.atlassian.net/browse/ANDROSDK-2336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/network/programstageworkinglist/ProgramStageWorkingListDTO.kt
```diff
@@ -55,8 +55,10 @@ internal data class ProgramStageWorkingListDTO(
         return ProgramStageWorkingList.builder()
             .applyBaseIdentifiableFields(this)
             .description(description)
-            .program(program?.toDomain())
-            .programStage(programStage?.toDomain())
+            .apply {
```
vgarciabnz: It might look like program and programStage are optional in domain, but they are mandatory. If any of those is null in the DTO, it will fail.
Maybe we can ignore the workinkList if any of them is null.

vgarciabnz (approved):

vgarciabnz (approved):

PR #2656, opened 2026-07-02
chore: [ANDROSDK-2339] migrate RepositoryScope to Kotlin

taridepaco:
Migrates `RepositoryScope` from a Java AutoValue class to a Kotlin data class using the `@ModelBuilder` annotation (domain models batch 20). It also adjusts `ModelBuilderProcessor` so that the builder backing properties generated for `internal` fields use `private` visibility instead of `protected`: this prevents an internal type from leaking through a public builder while avoiding name shadowing inside callers' `builder().apply { }` blocks.

It additionally fixes a null pointer bug in program stage section handling by moving the logic that assigns the parent program stage UID to the network layer instead of the handler.

Related task: [ANDROSDK-2339](https://dhis2.atlassian.net/browse/ANDROSDK-2339)

[ANDROSDK-2339]: https://dhis2.atlassian.net/browse/ANDROSDK-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/arch/repositories/scope/RepositoryScope.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.hisp.dhis.android.core.arch.repositories.scope.internal.RepositoryScopeComplexFilterItem
+import org.hisp.dhis.android.core.arch.repositories.scope.internal.RepositoryScopeFilterItem
+import org.hisp.dhis.android.core.arch.repositories.scope.internal.RepositoryScopeOrderByItem
+
+@ConsistentCopyVisibility
+@ModelBuilder
+data class RepositoryScope internal constructor(
+    val filters: List<RepositoryScopeFilterItem>,
+    val complexFilters: List<RepositoryScopeComplexFilterItem>,
+    internal val orderBy: List<RepositoryScopeOrderByItem>,
```
vgarciabnz, line 41: Why are these two properties internal? They were public before.

taridepaco: They were exposing an internal  classes (RepositoryScopeComplexFilterItem and RepositoryScopeOrderByItem)

vgarciabnz (approved):

PR #2657, opened 2026-07-02
feature: [ANDROSDK-2309] support isEmpty/isNotEmpty filter operators

andresmr:
Tracker filters (eventFilters, trackedEntityInstanceFilters and programStageWorkingLists) support an isEmpty/isNotEmpty operator for text values that the SDK previously did not handle. This PR adds full support for the operator, exposing a new isEmpty property on filters and building the corresponding SQL condition so the operator is honored in both offline and online queries.

Related task: [ANDROSDK-2309](https://dhis2.atlassian.net/browse/ANDROSDK-2309)

[ANDROSDK-2309]: https://dhis2.atlassian.net/browse/ANDROSDK-2309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/repositories/scope/internal/WhereClauseFromScopeBuilder.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/repositories/scope/internal/WhereClauseFromScopeBuilder.kt
@@ -36,7 +36,12 @@ internal class WhereClauseFromScopeBuilder(private val builder: WhereClauseBuild
             return "1"
         }
         for (item in scope.filters()) {
-            builder.appendKeyOperatorValue(item.key(), item.operator().sqlOperator, item.value() ?: "")
+            when (item.operator()) {
+                FilterItemOperator.NULL_OR_BLANK, FilterItemOperator.NOT_NULL_AND_NOT_BLANK ->
+                    builder.appendComplexQuery(item.operator().getSqlCondition(item.key()))
+
+                else -> builder.appendKeyOperatorValue(item.key(), item.operator().sqlOperator, item.value() ?: "")
+            }
         }
         for (item in scope.complexFilters()) {
             builder.appendComplexQuery(item.whereQuery())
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/hisp/dhis/android/core/arch/repositories/scope/internal/WhereClauseFromScopeBuilder.kt
```diff
@@ -36,7 +36,12 @@ internal class WhereClauseFromScopeBuilder(private val builder: WhereClauseBuild
             return "1"
         }
         for (item in scope.filters()) {
-            builder.appendKeyOperatorValue(item.key(), item.operator().sqlOperator, item.value())
+            when (item.operator()) {
```
vgarciabnz: This logic, is it not already defined in `FilterItemOperator.kt`?

core/src/main/java/org/hisp/dhis/android/core/trackedentity/search/TrackedEntityInstanceQueryOnlineHelper.kt
```diff
@@ -158,11 +158,17 @@ internal class TrackedEntityInstanceQueryOnlineHelper(
...
-                escapeChars(item.value())
+            return when (item.operator) {
+                FilterItemOperator.IN -> {
+                    val list = FilterOperatorsHelper.strToList(item.value()!!).map { escapeChars(it) }
+                    list.joinToString(";")
+                }
+
+                FilterItemOperator.NOT_NULL_AND_NOT_BLANK,
+                FilterItemOperator.NULL_OR_BLANK,
+                -> ""
```
vgarciabnz: I am not sure this is right. The api filter has the syntax:
`filter=w75KJ2mc4zz:null` or `filter=w75KJ2mc4zz:!null`. 

Also, I think the apiFilterOperator value is not correct for these two operators.

vgarciabnz (approved):
Good one! Added a commit with an adaptation for a failing test (just the assertion, not a functional error).

PR #2658, opened 2026-07-02
chore: [ANDROSDK-2340] migrate batch 21 models and remove AutoValue

taridepaco:
Migrate the last remaining AutoValue domain models to Kotlin data classes annotated with `@ModelBuilder` (Period, DatePeriod, ExpressionDimensionItem, DataApproval and DataSetCompleteRegistration), keeping the Java-style accessors and generated builders so the public API surface remains compatible. This completes the AutoValue-to-Kotlin migration: no domain model uses AutoValue anymore.

Since AutoValue was the only remaining annotation processor on kapt, this PR also removes the AutoValue dependency and the `kotlin-kapt` plugin from the build (code generation is now fully KSP), and updates CLAUDE.md and the architecture docs accordingly (including the outdated Dagger references, replaced by the current Koin setup).

Related task: [ANDROSDK-2340](https://dhis2.atlassian.net/browse/ANDROSDK-2340)

[ANDROSDK-2340]: https://dhis2.atlassian.net/browse/ANDROSDK-2340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2659, opened 2026-07-03
chore: [ANDROSDK-2241] update koin to 4.2.2 and migrate to the compiler plugin

taridepaco:
Updates Koin from 3.5.6 to 4.2.2 and replaces the KSP-based Koin Annotations processor with the new Koin Compiler Plugin (1.0.1), which requires Kotlin 2.3.20. This unblocks KSP2, so the `ksp.useKSP2=false` workaround is removed and KSP is updated to 2.3.9 (now only used by Room and the internal DAO processor). Module registration in `D2DIComponentFactory` moves from the generated `XxxDIModule().module` extensions to the plugin's typed `modules(KClass...)` API. The last classes registered through the classic DSL (`javaDIClasses`) are migrated to Kotlin and annotated with `@Singleton`, so the whole graph is now annotation-based.

With the DSL module gone, the plugin's compile-time dependency verification (`compileSafety`) is kept enabled for the main compilation. This required declaring explicit `includes` in the Network, Persistence and Wipe DI modules (real cross-module dependencies that the checker cannot resolve reliably otherwise), disabling `strictSafety`, and turning the checker off for test compilations, where it runs against a partial definition index and reports false positives. If spurious `KOIN-D001` errors ever show up in CI, the fix is either adding the corresponding `include` or disabling `compileSafety` until the plugin matures.

Note: this branch is stacked on top of #2658 (ANDROSDK-2340); the diff will narrow down to the Koin changes once that PR is merged.

Related task: [ANDROSDK-2241](https://dhis2.atlassian.net/browse/ANDROSDK-2241)

[ANDROSDK-2241]: https://dhis2.atlassian.net/browse/ANDROSDK-2241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vgarciabnz (approved):
I really don't understand why the WipeDIModule needs the declaration of the rest of the modules, but it fails if not added. We can check when @taridepaco is back.

PR #2661, opened 2026-07-07
chore: [ANDROSDK-2275] Adds plurals terminology to Program, ProgramStage and TrackedEntityType

ferdyrod:
Adds custom plurals terminology to Program object with `displayEnrollmentsLabel`, ` displayProgramStagesLabel`, `displayEventsLabel` now are available. To `ProgramStage` object `displayEventsLabel` is now available and the `TrackedEntityType` object `displayTrackedEntityTypesLabel` is now available.

Related to [ANDROSDK-2275](https://dhis2.atlassian.net/browse/ANDROSDK-2275)

[ANDROSDK-2275]: https://dhis2.atlassian.net/browse/ANDROSDK-2275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/network/program/ProgramDTO.kt
```diff
@@ -28,6 +28,7 @@
 
 package org.hisp.dhis.android.network.program
 
+import android.system.Os.access
```
vgarciabnz: This is not intentional, is it? I don't know why it was not detected by the checks...

core/src/main/java/org/hisp/dhis/android/core/program/ProgramStage.kt
```diff
@@ -82,6 +82,7 @@ data class ProgramStage(
     val validationStrategy: ValidationStrategy?,
     val displayProgramStageLabel: String?,
```
andresmr, line 83: why don't `displayProgramStagesLabel` also here?

core/src/main/java/org/hisp/dhis/android/network/programstage/ProgramStageDTO.kt
```diff
@@ -87,6 +87,7 @@ internal data class ProgramStageDTO(
     val displayProgramStageLabel: String?,
```
andresmr, line 87: same question here: why don't displayProgramStagesLabel also here?

core/src/main/java/org/hisp/dhis/android/network/program/ProgramDTO.kt
```diff
@@ -142,13 +146,16 @@ internal data class ProgramDTO(
             featureType(featureType?.let { FeatureType.valueOf(it) })
             accessLevel(accessLevel?.let { AccessLevel.valueOf(accessLevel) })
             displayEnrollmentLabel(displayEnrollmentLabel ?: enrollmentLabel)
+            displayEnrollmentsLabel(displayEnrollmentsLabel ?: enrollmentLabel)
```
vgarciabnz: I am not sure we should default to the singular value here, at least not in the SDK. It is important to tell the app that the plural translation is not defined so the app can decide what to to: the app can default to the singular translation or follow another approach.

What do you think? @ferdyrod @andresmr

core/src/main/java/org/hisp/dhis/android/core/program/ProgramStage.kt
vgarciabnz: It is not defined in the backend, it is only `displayProgramStageLabel` in ProgramStage.

core/src/main/java/org/hisp/dhis/android/network/program/ProgramDTO.kt
andresmr: yes, it makes sense to me.

ferdyrod: I agree, I'll change it

core/src/main/java/org/hisp/dhis/android/network/program/ProgramDTO.kt
ferdyrod: Don't know why this was added, will remove it

core/src/main/java/org/hisp/dhis/android/network/programstage/ProgramStageDTO.kt
```diff
@@ -122,6 +123,7 @@ internal data class ProgramStageDTO(
             validationStrategy(validationStrategy?.let { ValidationStrategy.valueOf(it) })
             displayProgramStageLabel(displayProgramStageLabel ?: programStageLabel)
             displayEventLabel(displayEventLabel ?: eventLabel)
+            displayEventsLabel(displayEventsLabel ?: eventLabel)
```
vgarciabnz: This one shouldn't default to singular

vgarciabnz (approved):

PR #2662, opened 2026-07-09
feature: [ANDROSDK-2345] build logout url

andresmr:
Adds a logout URL builder to the OAuth2 handler so apps can unregister the device after an offline login setup. `d2.userModule().oauth2Handler().buildLogoutUrl(serverUrl)` generates a state, persists it, and delegates to the network handler to build `$serverUrl/dhis-web-commons-security/logout.action?redirect_uri=$REDIRECT_URI&state=$state`, so the logout redirect can be correlated with the request. Unit tests cover both `OAuth2HandlerImpl` and `OAuth2NetworkHandlerImpl`.

Related task: [ANDROSDK-2345](https://dhis2.atlassian.net/browse/ANDROSDK-2345)

[ANDROSDK-2345]: https://dhis2.atlassian.net/browse/ANDROSDK-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
+++ b/core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
@@ -132,4 +132,10 @@ internal class OAuth2NetworkHandlerImpl(
             )
         }
     }
+
+    override fun buildLogoutUrl(serverUrl: String, state: String): String {
+        return "$serverUrl/dhis-web-commons-security/logout.action" +
+            "?redirect_uri=${OAuth2Config.DEFAULT_REDIRECT_URI}" +
+            "&state=$state"
+    }
 }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/hisp/dhis/android/network/oauth2/OAuth2NetworkHandlerImpl.kt
```diff
@@ -132,4 +132,10 @@ internal class OAuth2NetworkHandlerImpl(
             )
         }
     }
+
+    override fun buildLogoutUrl(serverUrl: String, state: String): String {
+        return "$serverUrl/dhis-web-commons-security/logout.action" +
+            "?redirect_uri=${OAuth2Config.DEFAULT_REDIRECT_URI}" +
```
vgarciabnz: The `OAuth2Config` class has a `redirectUri` property. Should we take this in the first place and, if it doesn't exist, default to the default one?

taridepaco (approved):

PR #2670, opened 2026-07-16
chore: [ANDROSDK-2357] migrate remaining Java tests to Kotlin - part 3

taridepaco:
Migrate the first half of the remaining Java test-support files in `core/src/sharedTest` to Kotlin (40 files), as described in the Jira task. These are the `*Samples` factory helpers (attribute, category, common, configuration, constant, dataelement, dataset, enrollment, imports, indicator, legendset, maintenance, note, option, organisationunit) plus the `RandomGeneratedInputStream` helper, all shared between the unit (`src/test`) and instrumented (`src/androidTest`) test source sets.

Each `*Samples` class becomes a Kotlin `object` whose factory methods keep their exact names and signatures and are annotated `@JvmStatic`, so every existing caller — Java and Kotlin, in both test source sets — keeps compiling unchanged with no consumer edits. The migration is behavior-preserving. Both `compileDebugUnitTest*` and `compileDebugAndroidTest*` were verified, and the 54 store integration tests that consume these samples were run on an emulator (1551 tests, all passing).

Related task: [ANDROSDK-2357](https://dhis2.atlassian.net/browse/ANDROSDK-2357)

[ANDROSDK-2357]: https://dhis2.atlassian.net/browse/ANDROSDK-2357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taridepaco (commented):
Reviewed by Tarí.

vgarciabnz (approved):

PR #2671, opened 2026-07-16
chore: [ANDROSDK-2358] migrate remaining Java tests to Kotlin - part 4

taridepaco:
Migrate the second half of the remaining Java test-support files in `core/src/sharedTest` to Kotlin (39 files), completing the `sharedTest` migration started in ANDROSDK-2357. These are the `*Samples` factory helpers for organisationunit, period, program, relationship, settings, systeminfo, trackedentity, user and validation, plus the `EventSync`/`TrackedEntityInstanceSync` samples, all shared between the unit (`src/test`) and instrumented (`src/androidTest`) test source sets.

Each `*Samples` class becomes a Kotlin `object` whose factory methods keep their exact names and signatures and are annotated `@JvmStatic`, and any `static final` constants become `const val`/`@JvmField`, so every existing caller — Java and Kotlin, in both test source sets — keeps compiling unchanged with no consumer edits. The migration is behavior-preserving. Both `compileDebugUnitTest*` and `compileDebugAndroidTest*` were verified, and the 72 store integration tests that consume these samples were run on an emulator (1967 tests, all passing).

Related task: [ANDROSDK-2358](https://dhis2.atlassian.net/browse/ANDROSDK-2358)

[ANDROSDK-2358]: https://dhis2.atlassian.net/browse/ANDROSDK-2358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taridepaco (commented):
Reviewed by Tarí.

vgarciabnz (approved):

PR #2674, opened 2026-07-20
fix: [ANDROSDK-2344] store cookies by host to avoid mixing sessions across servers

taridepaco:
## Summary

Fixes [ANDROSDK-2344](https://dhis2.atlassian.net/browse/ANDROSDK-2344): cookies were not well managed when calling external URLs through the SDK's `httpServiceClient()`. The HTTP client kept a single global cookie store, so cookies captured from one host (e.g. the DHIS2 instance) leaked into requests to a different host, and vice versa, corrupting the active session.

## Root cause

`CookieAuthenticatorHelper` held a single, host-agnostic `cookieMap`. Combined with the login-call detection (any request carrying an `Authorization` header is treated as a login), an external request would:
- clear the current DHIS2 session cookie, and
- store the external server's `set-cookie` into the shared map,

so the next call to the DHIS2 instance sent the wrong host's cookie.

## Change

Scope the cookie store by host (Option 2):
- `CookieAuthenticatorHelper` now keys cookies by host (`Map<host, Map<name, value>>`). Cookies are stored under the responding host and only attached to requests going to that same host.
- `isCookieDefined(host)` / `removeCookie(host)` now operate per host; the login branch in `ParentAuthenticatorPlugin` only clears the destination host's cookies.
- `PasswordAndCookieAuthenticator` passes the request host through.

The DHIS2 session-expiry fallback (redirect to login → retry with password) is unchanged, now scoped per host.

## Tests

`CookieAuthenticatorShould` rewritten to use `MockEngine` (real responses with a host), plus new cases:
- `not_mix_cookies_between_hosts`
- `remove_cookies_only_for_the_given_host`

`./gradlew :core:testDebugUnitTest` for the affected classes → passing.

[ANDROSDK-2344]: https://dhis2.atlassian.net/browse/ANDROSDK-2344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/api/authentication/internal/CookieAuthenticatorHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/api/authentication/internal/CookieAuthenticatorHelper.kt
@@ -38,9 +38,13 @@ internal class CookieAuthenticatorHelper {
     companion object {
         private const val COOKIE_KEY = "Cookie"
         private const val SET_COOKIE_KEY = "set-cookie"
+        private const val PATH_ATTRIBUTE = "path"
+        private const val DEFAULT_PATH = "/"
     }
 
-    private val cookieMapByHost = mutableMapOf<String, MutableMap<String, String>>()
+    private data class StoredCookie(val nameValue: String, val path: String)
+
+    private val cookieMapByHost = mutableMapOf<String, MutableMap<String, StoredCookie>>()
 
     fun storeCookieIfSentByServer(res: HttpResponse) {
         val host = res.call.request.url.host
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/hisp/dhis/android/core/arch/api/authentication/internal/CookieAuthenticatorHelper.kt
```diff
@@ -40,34 +40,40 @@ internal class CookieAuthenticatorHelper {
...
+    fun removeCookie(host: String) {
+        cookieMapByHost.remove(host)
     }
 
     fun addCookieHeader(requestBuilder: HttpRequestBuilder) {
-        requestBuilder.apply {
-            headers.remove(COOKIE_KEY)
-            header(COOKIE_KEY, cookieMap.values.joinToString("; "))
+        val host = requestBuilder.url.host
+        val hostCookies = cookieMapByHost[host]
```
vgarciabnz: I think this is not correct for instances under the same host name with different paths, although it shouldn't be a problem because it should reauthenticate if the cookie fails. After some testing, I have seen two things. I have updated the sample test in the JIRA ticket to reproduce it.

Let's suppose we have these two instances:
- https://play.im.dhis2.org/stable-2-41-9
- https://play.im.dhis2.org/stable-2-43-0-1

They are two different instances, but they have the same value for `host`. So the cookie is going to be reused. Each cookie includes extra information about the path
`JSESSIONID=4DD96301F71D2F5EC41DFD1D3BC012AB; Path=/stable-2-41-9; Secure; HttpOnly`.
And those cookies should only be used for this particular path.

Anyway, the SDK should sent the credentials again when it detects an invalid cookie. So it seems that this mechanism is broken in v43 (the redirection that we expected has changed). This verification is done in `PasswordAndCookieAuthenticator`.

vgarciabnz (changes requested):

vgarciabnz (approved):

PR #2680, opened 2026-07-23
feat: add FileCompressionHelper for image compression before upload

taridepaco:
## Summary

Adds `FileCompressionHelper`, a helper to compress image files before uploading them to the server with minimal quality loss. It iteratively downscales the image (preserving aspect ratio) until the encoded file drops below a target size, estimating the initial scale factor from the size ratio so a single pass is usually enough.

This is intended to eventually replace `FileResizerHelper` in the upload flow. It is **not wired into the upload path yet** — this PR only introduces the helper and its tests.

## Changes

- `FileCompressionHelper.kt` — new helper. Public entry `compressFile(file, targetSizeBytes = 600 KB)`; JPEG encoded at fixed quality, PNG size reduction via scaling only (lossless). Split into single-responsibility functions with safe bitmap recycling. Returns the original file if it cannot be decoded.
- `FileCompressionHelperShould.kt` — instrumented tests (require device/emulator): PNG/JPEG below target, aspect-ratio preservation, no upscaling when already below target, and passthrough for non-image files.
- `core/api/core.api` — regenerated via `./gradlew :core:apiDump` for the new public helper.

## Testing

- `./gradlew :core:apiDump` (compiles main + updates dump) — OK
- `./gradlew :core:detekt` — OK
- Instrumented tests require a connected device/emulator (`./gradlew :core:connectedDebugAndroidTest`).

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -1980,6 +1980,13 @@ public final class org/hisp/dhis/android/core/arch/helpers/DateUtils {
 	public static final fun getStartDate (Ljava/util/List;)Ljava/util/Date;
 }
 
+public final class org/hisp/dhis/android/core/arch/helpers/FileCompressionHelper {
+	public static final field INSTANCE Lorg/hisp/dhis/android/core/arch/helpers/FileCompressionHelper;
+	public static final fun compressFile (Ljava/io/File;)Ljava/io/File;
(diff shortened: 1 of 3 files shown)


taridepaco:
Reopening with the correct branch name and title convention (ANDROSDK-2349).

PR #2681, opened 2026-07-23
feature: [ANDROSDK-2349] implement default image compression

taridepaco:
Introduces `FileCompressionHelper`, the default image compression that is meant to replace the MEDIUM resolution used so far. Instead of targeting a fixed pixel dimension, it works against a size threshold in kB: it downscales the image preserving the aspect ratio until the encoded file drops below the threshold, estimating the initial scale factor from the size ratio so a single pass is usually enough. The threshold is hardcoded to 600 kB by default but exposed as a parameter so it can be adjusted easily.

The helper is not wired into the upload flow yet; this PR only adds the helper and its instrumented tests. JPEG is encoded at a fixed quality while PNG relies on scaling only (lossless), and a file that cannot be decoded as an image is returned unchanged. The public API dump was regenerated for the new helper.

Related task: [ANDROSDK-2349](https://dhis2.atlassian.net/browse/ANDROSDK-2349)

[ANDROSDK-2349]: https://dhis2.atlassian.net/browse/ANDROSDK-2349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileCompressionHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileCompressionHelper.kt
@@ -1,0 +1,138 @@
+/*
+ *  Copyright (c) 2004-2026, University of Oslo
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  Redistributions of source code must retain the above copyright notice, this
+ *  list of conditions and the following disclaimer.
+ *
+ *  Redistributions in binary form must reproduce the above copyright notice,
+ *  this list of conditions and the following disclaimer in the documentation
+ *  and/or other materials provided with the distribution.
+ *  Neither the name of the HISP project nor the names of its contributors may
+ *  be used to endorse or promote products derived from this software without
+ *  specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ *  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileCompressionHelper.kt
```diff
@@ -0,0 +1,138 @@
...
+            scale *= SCALE_STEP
+            iteration++
+        } while (iteration < MAX_ITERATIONS)
+    }
+
+    private fun writeScaledBitmap(source: Bitmap, scale: Float, format: CompressFormat, outputFile: File) {
+        val scaled = scaleBitmap(source, scale)
+        try {
+            FileOutputStream(outputFile).use { out ->
+                scaled.compress(format, JPEG_QUALITY, out)
```
vgarciabnz, line 102: The second parameter here can be interpreted differently ([here](https://developer.android.com/reference/android/graphics/Bitmap#compress(android.graphics.Bitmap.CompressFormat,%20int,%20java.io.OutputStream))), should we use a different quality depending on the compress format?

core/src/main/java/org/hisp/dhis/android/core/arch/helpers/FileCompressionHelper.kt
```diff
@@ -0,0 +1,138 @@
...
+            sqrt(targetSizeBytes.toFloat() / originalSizeBytes)
+        } else {
+            FULL_SIZE
+        }
+    }
+
+    private fun resolveCompressFormat(file: File): CompressFormat {
+        return when (FileResourceUtil.getExtension(file.name)?.lowercase()) {
+            "png" -> CompressFormat.PNG
+            else -> CompressFormat.JPEG
```
vgarciabnz, line 135: The server returns this error when an invalid image format is uploaded, which seems to mean that several format are supported, not just png or jpeg. Should we deal with them?

```
DataElement `IMAGE` is not valid: `File resource with uid 'LKWpelIIa0m' is using invalid image format webp. Valid formats are: JPG, JPEG 2000, tiff, bmp, PCX, gif, WBMP, PNG, RAW, JPEG, btiff, PNM, tif, TIFF, wbmp, jpeg, jpg, JPEG2000, BMP, pcx, GIF, png, raw, pnm, TIF, jpeg2000, jpeg 2000, BTIFF.`.
```

vgarciabnz (approved):

PR #2685, opened 2026-07-28
feature: [ANDROSDK-2366] persist image settings and add resource context

taridepaco:
Program and dataSet specific settings now carry an `imageSettings` map, keyed by item uid, holding the `UploadQuality` configured for that item (`DEFAULT` or `ORIGINAL`). The map is downloaded from the Android Settings web app, serialized into a new `imageSettings` column on `ProgramSetting` and `DataSetSetting` through a value class, and read back from it. Unknown qualities degrade to `DEFAULT` instead of failing, so a Settings web app offering a quality this SDK version does not know about cannot abort the whole settings download. The `@ModelBuilder` processor also needed to collect imports from nested type arguments for the generated builders to compile. This part covers [ANDROSDK-2348](https://dhis2.atlassian.net/browse/ANDROSDK-2348).

On top of that, `FileResourceCollectionRepository.suspendProcessAndAdd` takes a `ResourceContext` sealed class describing the origin of the file — a program image, a dataSet image or a plain file. Image contexts resolve their configured quality from the corresponding settings and are compressed unless `ORIGINAL` is set, while plain files are stored untouched. Note this branch is based on ANDROSDK-2349, so its compression helper shows up in the diff until that PR is merged.

Related task: [ANDROSDK-2366](https://dhis2.atlassian.net/browse/ANDROSDK-2366)

[ANDROSDK-2348]: https://dhis2.atlassian.net/browse/ANDROSDK-2348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ANDROSDK-2366]: https://dhis2.atlassian.net/browse/ANDROSDK-2366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/ResourceContext.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/arch/helpers/ResourceContext.kt
@@ -1,0 +1,49 @@
+/*
+ *  Copyright (c) 2004-2026, University of Oslo
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  Redistributions of source code must retain the above copyright notice, this
+ *  list of conditions and the following disclaimer.
+ *
+ *  Redistributions in binary form must reproduce the above copyright notice,
+ *  this list of conditions and the following disclaimer in the documentation
+ *  and/or other materials provided with the distribution.
+ *  Neither the name of the HISP project nor the names of its contributors may
+ *  be used to endorse or promote products derived from this software without
+ *  specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ *  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ *  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.helpers
+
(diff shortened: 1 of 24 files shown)


core/src/main/java/org/hisp/dhis/android/core/arch/helpers/ResourceContext.kt
```diff
@@ -0,0 +1,49 @@
...
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.helpers
+
+sealed class ResourceContext(val path: String) {
+    sealed class ImageContext(val filePath: String, resourceUid: String) : ResourceContext(filePath) {
+        data class ProgramImageContext(
+            val programUid: String,
+            val resourceUid: String,
```
vgarciabnz: This resource can be either a dataElement or an attribute, right? 

If we are going to use the same class for both, what is the property `attribute` for?

core/src/main/java/org/hisp/dhis/android/core/arch/helpers/ResourceContext.kt
```diff
@@ -0,0 +1,49 @@
...
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.hisp.dhis.android.core.arch.helpers
+
+sealed class ResourceContext(val path: String) {
```
vgarciabnz: The `path` property, is it needed? This context is going to be passed along a `File` object.

core/src/main/java/org/hisp/dhis/android/core/arch/helpers/ResourceContext.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        data class DatasetImageContext(
+            val datasetUid: String,
+            val resourceUid: String,
+            val attribute: String,
+            val imagePath: String,
+        ) : ImageContext(imagePath, resourceUid)
+    }
+
+    data class FileContext(val filePath: String) : ResourceContext(filePath)
```
vgarciabnz: Making a different context for Images and Files can be easier for the SDK, we can check with the app if this is ok (I guess it is). This is an information that we could have by checking the DB using the dataElement or attribute uid.

core/src/main/java/org/hisp/dhis/android/core/fileresource/FileResourceCollectionRepository.kt
```diff
@@ -117,6 +126,35 @@ class FileResourceCollectionRepository internal constructor(
         }
     }
 
+    suspend fun suspendProcessAndAdd(o: File, resourceContext: ResourceContext): String = withContext(dispatcher) {
+        val processedFile = when (resourceContext) {
+            is ResourceContext.ImageContext.ProgramImageContext -> compressImageIfNeeded(o, getQuality(resourceContext))
+            is ResourceContext.ImageContext.DatasetImageContext -> compressImageIfNeeded(o, getQuality(resourceContext))
```
vgarciabnz: I think it can simplified to a single line:
```
is ResourceContext.ImageContext -> compressImageIfNeeded(o, getQuality(resourceContext))
```

vgarciabnz (approved):

PR #2688, opened 2026-07-31
fix: [ANDROSDK-2342] download files of program attributes

taridepaco:
The tracker API needs the program uid to resolve the file of a tracked entity attribute that is assigned to a program. The SDK was not sending it, so images and files of program attributes were never downloaded, while the ones assigned to the tracked entity type worked fine. File resources can now be downloaded atomically with the tracker data, through `TrackedEntityInstanceDownloader.downloadFileResources()` and `EventDownloader.downloadFileResources()`: downloading them in the same iteration that persists each batch is what keeps the program of every item known.

When there is no download context, which is the case of the standalone `FileResourceDownloader`, the program is inferred from the local metadata: the programs the attribute is assigned to, intersected with the enrollments of the tracked entity. That covers the usual case but can pick the wrong program when an attribute belongs to several, so the tracker filters of that downloader are now deprecated with that caveat. Only tracked entities and events are atomised here; aggregated data values and custom icons still download separately.

Related task: [ANDROSDK-2342](https://dhis2.atlassian.net/browse/ANDROSDK-2342)

[ANDROSDK-2342]: https://dhis2.atlassian.net/browse/ANDROSDK-2342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCallHelper.kt
+++ b/core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCallHelper.kt
@@ -103,7 +103,7 @@ internal class FileResourceDownloadCallHelper(
         }
 
         val attributeValues = trackedEntityAttributeValueStore.selectWhere(attributeValuesWhereClauseBuilder.build())
-        val resolveProgram = buildAttributeProgramResolver(attributeValues, params.contextProgramUid)
+        val resolveProgram = buildAttributeProgramResolver(attributeValues, params.contextProgram())
 
         return attributeValues.map { av ->
             val type = trackedEntityAttributes.find { it.uid() == av.trackedEntityAttribute() }!!.valueType()!!
@@ -112,13 +112,10 @@ internal class FileResourceDownloadCallHelper(
     }
 
     /**
-     * Builds a function that resolves the program to use when downloading the file of an attribute value. The tracker
-     * API requires the program uid for attributes assigned to a program: without it the file is not resolved.
+     * Builds a function that resolves the program to send when downloading the file of an attribute value. The
+     * tracker API requires it for attributes assigned to a program: without it the file is not resolved.
      *
-     * When the file resources are downloaded along with the tracker data, the program is known and is preferred. In
-     * any other case it is inferred from the local metadata: the programs the attribute is assigned to, intersected
-     * with the programs the tracked entity is enrolled in. Attributes that are not assigned to any program resolve to
-     * null, as the endpoint works without the parameter for tracked entity type attributes.
+     * The metadata both decisions are based on is read once for the whole batch, not once per value.
      */
     private suspend fun buildAttributeProgramResolver(
         attributeValues: List<TrackedEntityAttributeValue>,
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadParams.kt
```diff
@@ -43,6 +43,15 @@ internal data class FileResourceDownloadParams(
...
     val maxContentLength: Int? = null,
+    /**
+     * Program the file resources are being downloaded for. Unlike [programUids], it is not a filter: it is the
+     * program used to build the tracked entity attribute file / image endpoints, which the tracker API requires for
+     * attributes assigned to a program.
+     *
+     * It is only known when the file resources are downloaded as part of a tracker data download. Otherwise it is
+     * null and the program is inferred from the local metadata, which might be ambiguous.
+     */
+    val contextProgramUid: String? = null,
```
vgarciabnz: Is it not weird that this attribute has such a long comment compared to the rest of attributes?

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCallHelper.kt
```diff
@@ -96,11 +102,65 @@ internal class FileResourceDownloadCallHelper(
...
+
+        val programsByTrackedEntity = enrollmentStore.selectWhere(enrollmentsWhereClause)
+            .groupBy({ it.trackedEntityInstance() }, { it.program() })
+            .mapValues { (_, programs) -> programs.filterNotNull().toSet() }
+
+        return { value ->
+            val candidates = programsByAttribute[value.trackedEntityAttribute()].orEmpty()
+            when {
+                candidates.isEmpty() -> null
+                contextProgramUid != null && candidates.contains(contextProgramUid) -> contextProgramUid
```
vgarciabnz: Why would we not use the `contextProgramUid` if defined? Shouldn't it have priority over the rest?

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadParams.kt
vgarciabnz: Also, I think if it could make sense that this programUid was a particular case of `programUids` with a single value. I mean, if we could reuse the other property. In this way, this program will be taken into account in the program granular sync using the deprecated downloader because it passes the `programUid` as a filter.

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadCallHelper.kt
taridepaco: There are three cases: 
1. The attribute is not assigned to any program → no `program` is sent. The endpoint resolves tracked entity type attributes without it.
2. The attribute is assigned to the program the data was downloaded for → that one wins over any inference. It is the only candidate we know is right.
3. The attribute is assigned to some *other* program → the context does not apply to this attribute, so we infer: a program the attribute belongs to and the tracked entity is enrolled in.

I'm going to refactor the resolver a bit to make it clearer

core/src/main/java/org/hisp/dhis/android/core/fileresource/internal/FileResourceDownloadParams.kt
taridepaco: they mean different things:

- `programUids` is a filter. It is consumed by `getTrackedEntityWhereClauseFromParams()` and `getEventWhereClauseFromParams()` to narrow down which pending values are selected.
- `contextProgramUid` is not a filter. It never touches the query; it only ends up in the request URL as `?program=`.

If the atomic download passed `programUids = listOf(program)` instead, the selection of missing values would gain an extra `uid IN (SELECT trackedEntityInstance...

Nevertheless, if there is a single program in the filter, it can be used as a fallback for the context program, simplifying the resolver flow.

vgarciabnz (approved):

PR #2693, opened 2026-08-06
Updating Dependency-Track action script

michael-markevich:
The updated script uses a new variable for the Dependency-Track URL and checks the response status by returning the processing token.

### DIFF
--- a/.github/workflows/generate-and-upload-bom.yml
+++ b/.github/workflows/generate-and-upload-bom.yml
@@ -24,11 +24,22 @@ jobs:
         run: ./gradlew cyclonedxBom
       - name: Upload SBOM to DependencyTrack
         env:
-          DEPENDENCY_TRACK_API: 'https://dt.security.dhis2.org/api/v1/bom'
+          DT_API_URL: ${{ vars.DEPENDENCYTRACK_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-and-upload-bom.yml
```diff
@@ -24,11 +24,22 @@ jobs:
...
+          response=$(curl -sS -X POST "$DT_API_URL" \
               --fail-with-body \
-              -H "Content-Type: multipart/form-data" \
-              -H "X-Api-Key: ${{ secrets.DEPENDENCYTRACK_APIKEY }}" \
-              -F "project=189ca834-ad84-4d54-ac98-c3d0526f56cc" \
-              -F "bom=@build/reports/bom.json"
+              -H "X-Api-Key: $DT_API_KEY" \
+              -F "autoCreate=true" \
+              -F "projectName=$PROJECT_NAME" \
+              -F "projectVersion=$PROJECT_VERSION" \
```
vgarciabnz, line 37: This is using `github.ref_name`, which is the branch name (not the version). I assume it is ok, isn't it?

vgarciabnz (approved):

.github/workflows/generate-and-upload-bom.yml
michael-markevich: Yes, that's correct and intended.

PR #2696, opened 2026-08-13
feature: [ANDROSDK-2372] identify expired or missing OAuth2 tokens

taridepaco:
The SDK can now tell the app when an OAuth2 session is no longer usable. A new `D2ErrorCode.OAUTH2_NO_VALID_TOKEN` is thrown when there is no OAuth2 state, no access token, no refresh token, or when the server explicitly rejects the refresh token, and also on login for an account registered as OAUTH2 that has no tokens on this device (for example, after importing a database). Token rotation is now centralized in `OAuth2TokenRefresher.rotate`, which serializes concurrent refreshes with a mutex and returns a `RefreshResult` distinguishing success, a retryable failure, and an unrecoverable one.

The distinction matters because the previous implementation logged the user out on any refresh failure, including transient ones: going offline or hitting a server error closed the session. Now only a 400/401 from the token endpoint discards the stored tokens; anything else leaves the credentials untouched and the caller sees the original 401 so it can retry later. Proactive refreshing based on `expiresAt` was dropped along with `OAuth2State.needsTokenRefresh()` — refresh is driven by the actual 401 response instead, which also removes the reliance on device clock accuracy.

Unified the OpenID Connect and OAuth2 re-login flows. A second login for either type is now local: it imports the pending database using the PIN as the offline code, or opens the existing one offline, without contacting the server. The PIN plays for token-based accounts the role the password plays for traditional ones.

The OpenID authenticator and token refresher were aligned with their OAuth2 counterparts, which the unified flow depends on: a refresh that fails for a transient reason (offline, provider outage) now keeps the session open instead of logging the user out, and only a refresh token the provider explicitly rejects is treated as definitive — reported through the new `OPEN_ID_CONNECT_NO_VALID_TOKEN` error code and clearing the stored state so the next login is deterministic. A missing state is also reported as that error instead of crashing.

Related task: [ANDROSDK-2372](https://dhis2.atlassian.net/browse/ANDROSDK-2372)

[ANDROSDK-2372]: https://dhis2.atlassian.net/browse/ANDROSDK-2372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andresmr (dismissed):

andresmr (dismissed):

core/src/main/java/org/hisp/dhis/android/core/user/internal/LogInCall.kt
```diff
@@ -82,13 +85,24 @@ internal class LogInCall(
...
+            isOAuth2AccountWithoutState(trimmedServerUrl, username) ->
+                throw exceptions.oauth2NoValidTokenError("This account has no OAuth2 tokens on this device")
             else -> {
                 exceptions.throwExceptionIfPasswordNull(passwordOrPin)
                 logInWithPassword(trimmedServerUrl, username, passwordOrPin)
             }
         }
     }
 
+    private fun isOAuth2AccountWithoutState(serverUrl: String, username: String): Boolean {
```
vgarciabnz: This method's name is a bit confusing, it only checks that the account is Oauth2, right?

core/src/main/java/org/hisp/dhis/android/core/user/internal/LogInCall.kt
```diff
@@ -82,13 +85,24 @@ internal class LogInCall(
                 logInWithOAuth2State(trimmedServerUrl, username, passwordOrPin, oauth2State)
             openIdState != null ->
                 logInWithOpenIdState(trimmedServerUrl, username, passwordOrPin, openIdState)
+            isOAuth2AccountWithoutState(trimmedServerUrl, username) ->
```
vgarciabnz: In the description says that this is the situation after importing a DB, but it shouldn't throw an exception here. The SDK should be able to load the DB and work offline even when there are no tokens. 
This is going to be supported in 3.6, I don't know if it makes sense to make it clearer here to avoid confusions.

core/src/main/java/org/hisp/dhis/android/core/user/internal/LogInCall.kt
```diff
@@ -77,26 +81,86 @@ internal class LogInCall(
...
+    private suspend fun logInWithOpenId(
+        serverUrl: String,
+        username: String,
+        pin: String?,
+        state: AuthState?,
+    ): User {
+        return if (state == null) {
+            // No tokens for this account on this device: an imported database, or a state that was
+            // wiped. There is nothing to authenticate with, so the database is opened offline.
+            importOrLogInOffline(
```
vgarciabnz: I was wondering if we could get openId offline login for free, but I think we have never truly talked about it. Maybe something we could consider.

core/src/main/java/org/hisp/dhis/android/core/user/internal/LogInCall.kt
```diff
@@ -114,7 +178,7 @@ internal class LogInCall(
                 networkHandler.authenticate("Bearer $token")
             }.getOrThrow()
             openIDConnectStateSecureStore.set(serverUrl, username, state)
-            loginOnline(user, credentials)
+            loginOnline(user, credentials, verifyPin = true)
```
vgarciabnz: Should we verify pin in OpenId flow?

vgarciabnz (approved):

PR #2700, opened 2026-08-20
chore: [ANDROSDK-2372] Set proper authorization type from Credentials when creating a fresh db

taridepaco:
Opening the database of an existing account no longer overwrites its authorization type with BASIC. `loadExistingChangingEncryptionIfRequired` now receives the type the login has just established and gives it precedence over the recorded one, so `MultiUserDatabaseManager.loadExistingKeepingEncryption` takes it as a parameter and `LogInDatabaseManager` receives the credentials it comes from.                      
                                                                                                                                                                                                                                                                                                                                                                                                                                 
Accounts created before the authorization type existed have it null, and defaulting them to BASIC labelled a token-based account as a password one for good. It went unnoticed because a stored OAuth2 or OpenID state takes precedence when the type is resolved at login; but once that state is discarded — after the provider rejects the refresh token — the login was routed to the password flow with no way out. The type is now corrected on the first login after upgrading, which also covers accounts opened offline.

Related task: [ANDROSDK-2372](https://dhis2.atlassian.net/browse/ANDROSDK-2372)

[ANDROSDK-2372]: https://dhis2.atlassian.net/browse/ANDROSDK-2372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/org/hisp/dhis/android/persistence/db/access/RoomMultiUserDatabaseManager.kt
+++ b/core/src/main/java/org/hisp/dhis/android/persistence/db/access/RoomMultiUserDatabaseManager.kt
@@ -137,7 +137,7 @@ internal class RoomMultiUserDatabaseManager(
     override suspend fun loadExistingKeepingEncryption(
         serverUrl: String,
         username: String,
-        authorizationType: AuthorizationType?,
+        authorizationType: AuthorizationType,
     ): Boolean {
         return loadExistingChangingEncryptionIfRequired(
             serverUrl,
@@ -198,7 +198,7 @@ internal class RoomMultiUserDatabaseManager(
         username: String,
         encryptionExtractor: (config: DatabaseAccount) -> Boolean,
         alsoOpenWhenEncryptionDoesntChange: Boolean,
-        authorizationType: AuthorizationType?,
+        authorizationType: AuthorizationType,
     ): Boolean {
         val existingAccount = getAccount(serverUrl, username) ?: return false
         val encrypt = encryptionExtractor(existingAccount)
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/hisp/dhis/android/persistence/db/access/RoomMultiUserDatabaseManager.kt
```diff
@@ -200,7 +209,9 @@ internal class RoomMultiUserDatabaseManager(
                 username,
                 encrypt,
                 loginConfig = existingAccount.loginConfig(),
-                authorizationType = existingAccount.authorizationType ?: AuthorizationType.BASIC,
+                authorizationType = authorizationType
+                    ?: existingAccount.authorizationType
+                    ?: AuthorizationType.BASIC,
```
andresmr: is possible to arrive here without `authorizationType` or `existingAccount.authorizationType` ?

core/src/main/java/org/hisp/dhis/android/persistence/db/access/RoomMultiUserDatabaseManager.kt
```diff
@@ -200,7 +209,9 @@ internal class RoomMultiUserDatabaseManager(
                 username,
                 encrypt,
                 loginConfig = existingAccount.loginConfig(),
-                authorizationType = existingAccount.authorizationType ?: AuthorizationType.BASIC,
+                authorizationType = authorizationType
+                    ?: existingAccount.authorizationType
```
vgarciabnz: I just wonder whether the existing authorizationType should take precedence over the passed authorizationType. My feeling is that they should NEVER be different if both of them are defined, so the order wouldn't matter.
Is there any case where they could be different?

andresmr (approved):

PR #2703, opened 2026-08-24
chore: [ANDROSDK-2059] replace MD5 password hashing with PBKDF2

taridepaco:
`UserHelper.md5()` is no longer used to store password hashes. Credentials are now hashed with PBKDF2-HMAC-SHA256 (210,000 iterations, 16-byte random salt) through a new internal `PasswordHasher`, and `Credentials.getHash()` is replaced by `newPasswordHash()` for the write path and `matches()` for the read path, so a salted hash is never compared for equality.

Existing accounts migrate transparently rather than being logged out. Hashes are stored in a self-describing `$pbkdf2-sha256$i=<iterations>$<salt>$<hash>` format: legacy MD5 values are still verified (verify-only, never produced again) and are rewritten in place on the next successful login, while the plaintext secret is still at hand. The same mechanism allows raising the iteration count in a later release without invalidating anything already stored. `UserHelper.md5` is kept as `@Deprecated` because it is published API and the legacy verification path still needs it; removal is left for the next major. Note that `PBKDF2WithHmacSHA256` is only available from API 26, so devices below that fall back to `PBKDF2WithHmacSHA1`, which is recorded in the stored value and upgraded when the account is next verified on a capable device.

Related task: [ANDROSDK-2059](https://dhis2.atlassian.net/browse/ANDROSDK-2059)

[ANDROSDK-2059]: https://dhis2.atlassian.net/browse/ANDROSDK-2059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/arch/storage/internal/Credentials.kt
```diff
@@ -45,8 +44,27 @@ internal data class Credentials(
...
         get() = password ?: pin
 
-    fun getHash(): String? {
-        return passwordOrPin?.let { UserHelper.md5(username, it) }
+    /**
+     * Derives a new hash for the stored secret, to be persisted in the AuthenticatedUser table.
+     * The result is salted and therefore different on every call, so it must never be compared:
+     * use [matches] to verify a secret against an already stored hash.
+     */
+    fun newPasswordHash(): String? {
```
vgarciabnz, line 52: I would avoid adding the `new` word in a method name because after some versions it won't be new anymore.
Why don't keep the same name as before? This method is replacing it.

core/src/main/java/org/hisp/dhis/android/core/arch/storage/internal/Credentials.kt
```diff
@@ -45,8 +44,27 @@ internal data class Credentials(
...
+
+    /**
+     * Verifies the stored secret against [storedHash], which may be in either the current or the
+     * legacy MD5 format. Accounts without a secret (token based accounts with no PIN) are expected
+     * to have no stored hash either.
+     */
+    fun matches(storedHash: String?): HashVerification {
+        val secret = passwordOrPin
+        return when {
+            secret == null && storedHash == null -> HashVerification.Match(needsUpgrade = false)
```
vgarciabnz, line 64: In case the existing hash was null, I am not sure if we should let the user in by providing a null value. If the hash is null, it means that the user didn't set up a PIN code.

core/src/main/java/org/hisp/dhis/android/core/arch/storage/internal/PasswordHasher.kt
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @Suppress("ReturnCount")
+    private fun verifyPbkdf2(secret: String, storedHash: String): HashVerification {
+        val parsed = parse(storedHash) ?: return HashVerification.Mismatch
+        val candidate = try {
+            deriveKey(parsed.algorithm, secret, parsed.salt, parsed.iterations)
+        } catch (_: NoSuchAlgorithmException) {
+            // The hash was written on a device running a newer Android version. It cannot be
+            // verified here, so the account has to be authenticated online again.
```
vgarciabnz, line 131: We would have to check if this restriction is important when exporting/importing a DB from a newer Android version. In version 3.6 we should move to use a different password for DB encryption and then force the user the do an online login in case they want to synchronize. I think it won't be so important if we implement this workflow, but it might be important if we keep the current one.

PR #2713, opened 2026-09-15
chore: [ANDROSDK-2384] review dataSet download flow

taridepaco:
The `/api/dataValueSets` request was not constraining the attributeOptionCombo, which forces a much slower query plan on the server. It now sends the attributeOptionCombos of the dataSet being downloaded, resolved from the local metadata. The `paging=false` parameter is also dropped: it is not a supported parameter of this endpoint and is ignored by the backend. On top of that, the number of future periods was stored with a different default than the one used to build the bundle, so the comparison in `AggregatedDataSyncLastUpdatedCalculator` never matched and dataSets without `openFuturePeriods` were fully re-downloaded on every sync instead of incrementally.

The attributeOptionCombos are only added while they fit in the maximum url length, falling back to the previous unfiltered query otherwise, since sending a truncated list would silently skip the data values of the omitted combos. Partitioning the request to keep it within both the url and the payload limits is handled in ANDROSDK-2203.

Related task: [ANDROSDK-2384](https://dhis2.atlassian.net/browse/ANDROSDK-2384)

[ANDROSDK-2384]: https://dhis2.atlassian.net/browse/ANDROSDK-2384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/java/org/hisp/dhis/android/core/dataset/DataSet.kt
```diff
@@ -110,6 +110,11 @@ data class DataSet(
     class Builder : DataSetBuilder()
 
     companion object {
+        internal const val DEFAULT_FUTURE_PERIODS = 1
```
andresmr, line 113: Why default future periods is 1?

andresmr (approved):

core/src/main/java/org/hisp/dhis/android/core/dataset/DataSet.kt
taridepaco: There was an inconsistency before, that led to always download 1 future period if the openFuturePeriods parameter was not set. This fix doesn't change the behaviour (setting it at 1) so we may decide to change it to 0, thus changing what was happening until now.

